PR #41, opened 2026-08-08
feat: compress blockIds for non-contiguous T2/T3 distillation

ranxianglei:
## Problem

The model frequently fails T2/T3 distillation calls. Root cause: the `compress` schema only accepts a **contiguous** `startId..endId` range. When the model wants to distill a set of non-contiguous blocks, it can't express "only these blocks" — a span would consume everything in between. So it dumps the block list into the `summary` text and omits `startId`/`endId`, which schema validation then rejects.

## Fix

Add an explicit `blockIds: [...]` option to `applyCompression` ranges — consume **exactly** the listed blocks into one higher-tier block, leaving any raw messages between them visible. This is the selective distillation the contiguous span cannot express.

### Changes
- `resolveBlockListBoundaries` (`src/compress.ts`): validates each id (must be `bN`, exist, active, **same tier**), unions their `effectiveMessageIds` as coverage, sets `boundaryKind="block"`.
- `resolveRangeSpec`: dispatches `blockIds` vs `resolveBoundaries`; `blockIds` takes precedence when non-empty.
- Overlap detection + `minCompressRange` pre-pass route through the same resolver, so `blockIds` ranges interoperate with message ranges in a batch.
- T2/T3 nudge (`src/nudge-text.ts`) now emits a `blockIds` example built from the actual target blocks, so the model gets the right call shape directly.

### Backward compatible
The contiguous `startRef`/`endRef` path is unchanged. Existing callers/tests unaffected.

### Tests
9 new tests in `tests/compress-blockids.test.ts`: T1→T2 distill (intervening msg stays visible), T2→T3 distill, precedence over startRef/endRef, batch with message range, shared-block overlap rejection, and error cases (invalid id, nonexistent, inactive/consumed, mixed tier). Full suite: **221/221 pass**, typecheck + build clean.

### Validation rules
- Each id must be a valid `bN`.
- Each must exist and be active.
- All must be the same tier (distill one tier at a time).
- No overlap with other ranges in the batch.

## Downstream
Once merged + published, `billion-context-pi` will bump its `acp-kernel` pin and add `blockIds` to its `compress` tool schema (separate PR).

🤔 This PR was created by an AI agent. **Do not merge without human review.** Per AGENTS.md, PR merges are human-only.

### DIFF
--- a/src/compress.ts
+++ b/src/compress.ts
@@ -9,7 +9,12 @@ import {
 } from "./state.js";
 import { defaultCountTokens } from "./tokenize.js";
 import { validateConfig } from "./config.js";
-import { resolveBoundaries, earliestIndexOfIds } from "./boundaries.js";
+import {
+  resolveBoundaries,
+  earliestIndexOfIds,
+  parseBoundary,
+  type ResolvedRange,
+} from "./boundaries.js";
 import { truncateLargeToolOutputs } from "./truncate-tools.js";
 import { hideConsumedCompressCalls } from "./hide-consumed.js";
 import { applyMessageFilters, listMessageFilters } from "./filter/index.js";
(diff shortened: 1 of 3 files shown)


ranxianglei:
连续的有助于减少碎片 应该连续才对

Issue #42, opened 2026-08-09
Errors: content: range (m15930..m16063) overlaps (m16062..m16168). Overlapping ranges cannot be compressed in the same batch.


ranxianglei:
Errors: content: range (m15930..m16063) overlaps (m16062..m16168). Overlapping ranges cannot be compressed in the same batch.

ranxianglei:
应该允许这个发生 手动帮引擎调节对位置

PR #55, opened 2026-08-10
fix: overlap warn+skip instead of aborting the whole batch (ISSUE-42)

ranxianglei:
## 背景 / Background

When two compress ranges in a single batch overlap, the previous code returned a hard error and discarded the **entire** batch (`blocksCreated: 0`). Overlap is minor and should not block compression.

Refs: acp-kernel#42, dog/billion-context-pi#21

## 改动 / Change

In `src/compress.ts` `applyCompression()` batch overlap pre-check:

- **Before**: detect overlap → `return { errors: [...overlaps...], blocksCreated: 0 }` (whole batch aborted).
- **After**: detect overlap → push a **warning**, add the later-overlapping range to `skipSpecs`, and continue. Earliest range (by lowest message index) wins; disjoint ranges still compress normally.

Also updated the `minCompressRange` threshold loop to skip discarded ranges so they don't inflate the aggregate char count.

## 行为 / Behavior

- Overlapping range → warning `Skipped range (X..Y) — overlaps an earlier range in the batch; the earlier range takes precedence. Keep ranges disjoint.` + that range is skipped.
- Earliest range still compressed.
- Disjoint ranges in the same batch still compressed.

## 测试 / Tests

- Updated `batch: overlapping ranges are rejected` → `batch: overlapping ranges warn and skip (earliest wins) instead of failing`: asserts `errors.length === 0`, `blocksCreated === 1`, warning present.
- New `batch: disjoint range still compresses when two others overlap`: 2 overlaps + 1 disjoint → `blocksCreated === 2`, 1 warning.
- Existing `non-overlapping ranges all succeed` unchanged.

Full suite: **213/213 pass**, typecheck + build clean.

## billion-context-pi compatibility

Validated per AGENTS.md §5 local overlay: overlaid new `acp-kernel/dist` → billion-context-pi `node_modules`; **158/158 tests pass**, build clean, new warning string confirmed bundled into billion-context-pi `dist`. (No version bump committed yet — acp-kernel 0.0.18 must publish first per cross-repo release order.)

### DIFF
--- a/src/compress.ts
+++ b/src/compress.ts
@@ -142,30 +142,28 @@ export function createCore(ports: Ports = {}): CompressionCore {
       const bMin = b.indices.length > 0 ? Math.min(...b.indices) : Infinity;
       return aMin - bMin;
     });
-    for (let i = 1; i < sortedRanges.length; i++) {
-      const prev = sortedRanges[i - 1]!;
-      const curr = sortedRanges[i]!;
-      const prevMax = prev.indices.length > 0 ? Math.max(...prev.indices) : -1;
-      const currMin = curr.indices.length > 0 ? Math.min(...curr.indices) : -1;
-      if (prevMax >= currMin && prevMax >= 0) {
-        return {
-          state: input.state,
-          result: {
-            blocksCreated: 0,
-            tokensCompressed: 0,
-            errors: [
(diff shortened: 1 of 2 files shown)


ranxianglei:
close https://github.com/ranxianglei/acp-kernel/issues/42

PR #91, opened 2026-08-20
fix: align nudge summary token counts with tokenizer

### LINKED ISSUE #45: nudge text shows summary tokens estimated via length/4, inconsistent with the CJK-aware tokenizer
## Context

v0.0.17, `b9f9e99`. The tier-distillation nudge renders a per-target-block line like:

```
  b1  37 msgs  56K→14K  "topic"
```

where `14K` is supposed to be the summary's token count. But `formatTierTargetBlocks` computes it with a raw chars/4 estimate:

```ts
// src/nudge-text.ts:39
const summaryTokens = Math.ceil((b.summary ?? "").length / 4);
```

## Why this is wrong

The rest of the engine uses an injected `countTokens`, and ships `defaultCountTokens` (`src/tokenize.ts:5-13`) which is **explicitly CJK-aware**: CJK characters count ~1:1 instead of being divided by 4, because chars/4 badly underestimates CJK content.

```
// src/tokenize.ts
const cjk = text.match(/[\u4e00-\u9fff\u3040-\u30ff\uac00-\ud7af]/g);
const cjkCount = cjk?.length ?? 0;
return cjkCount + Math.ceil((text.length - cjkCount) / 4);
```

For a summary written in Chinese/Japanese/Korean, `length/4` in `formatTierTargetBlocks` will **severely understate** the token count the model sees in the nudge. The model uses those numbers to judge compression payoff ("`56K→14K`"), so an understated `→14K` misrepresents the gain. Meanwhile `pendingByTier` (`src/compress.ts:829-830`) computes the *same* quantity with the real tokenizer for the gate decision — so the gate and the displayed number disagree.

## Root cause

`renderNudgeText(decision: NudgeDecision)` only receives the `NudgeDecision`; it has no access to `countTokens`. So it falls back to the crude estimate.

## Suggested fix

Two options:

- **(preferred) Precompute.** Have the kernel fill in a `summaryTokens` field on each entry of `NudgeDecision.tierTargetBlocks` (or on a parallel `tierTargetBlockStats[]`) using the injected `countTokens` it already holds. The renderer then just reads the number — rendering stays dependency-free.
- **(alternative) Pass tokenizer.** Add an optional `countTokens?: (t: string) => number` parameter to `renderNudgeText`, defaulting to `defaultCountTokens`. Keeps the type import but pushes the dependency to the call site.

Option 1 is cleaner: it preserves the "renderer takes plain data" contract and guarantees the displayed number matches the one the gate used.

## Severity

Medium. Not a crash, but the nudge is the primary signal the model uses to decide *whether* and *how much* to compress. A systematically wrong "tokens after compression" figure for CJK content steers the model toward bad compression decisions (under-compressing because the gain looks smaller than it is).

Beatrice0377:
## Problem

Tier-distillation nudges display each target block's summary token count, but the renderer estimated that value with `summary.length / 4`.

The decision layer already uses the injected `countTokens` function for the same summaries. This makes the displayed compression payoff diverge from the value used for tier arbitration, especially for CJK text and custom tokenizers.

## Root Cause

`pendingByTier` had access to the active `countTokens`, but `NudgeDecision` only carried the target `CompressionBlock`s.

The per-block token counts were therefore lost at the decision boundary, forcing `renderNudgeText` to estimate them again with `length / 4`.

## Fix

- Precompute per-target-block summary token counts in the decision layer using the active injected `countTokens`.
- Derive `pendingT2` / `pendingT3` from those same precomputed values so arbitration and rendering share one token-counting source.
- Carry the derived values on `NudgeDecision` as `tierTargetBlockStats`.
- Make the renderer consume those values instead of re-tokenizing normal kernel-produced decisions.
- Preserve the previous `length / 4` behavior as a compatibility fallback for manually constructed legacy decisions that do not provide stats.

## Scope / Non-goals

- `CompressionBlock` is unchanged; derived token counts are not persisted into runtime state.
- `renderNudgeText(decision, prompts?)` keeps its existing signature.
- No dependency, version, lockfile, workflow, or documentation changes.
- No unrelated refactors.

## Tests

Added regression coverage for:

- CJK summaries using the default CJK-aware tokenizer.
- A custom injected tokenizer that differs from both `length / 4` and `defaultCountTokens`.
- Renderer use of precomputed per-block token counts.
- Legacy decisions without stats.
- ASCII output compatibility.

Validation:

- Targeted tests: 47/47 passed.
- Full suite: 390/390 passed.
- `npm run typecheck`: passed.
- `npm run build`: passed.
- `git diff --check`: passed.

## Compatibility

The new `NudgeDecision.tierTargetBlockStats` field is optional.

Existing manually constructed decisions without the field continue to use the previous renderer estimate, while normal kernel-produced tier-2/tier-3 decisions provide tokenizer-consistent stats.

## Limitations

Legacy/manual decisions that omit `tierTargetBlockStats` intentionally retain the previous fallback estimate for backward compatibility.

Fixes #45

### DIFF
--- a/src/compress.ts
+++ b/src/compress.ts
@@ -47,6 +47,7 @@ import type {
   ProcessTurnResult,
   Recommendation,
   StatusReport,
+  TierTargetBlockStat,
 } from "./types.js";
 
 export interface Ports {
@@ -913,24 +914,40 @@ function resolveAdaptiveGrowth(
  *  atomically reject — see CompressibleRange.chars); T2 = total summary
  *  tokens of all active tier-1 blocks; T3 = total summary tokens of all
  *  active tier-2 blocks. */
+interface TierPending {
+  pending: number;
+  targetBlocks: CompressionBlock[];
+  targetBlockStats: TierTargetBlockStat[];
+}
+
 function pendingByTier(
   state: CompressionState,
   recommendation: Recommendation | undefined,
   countTokens: (t: string) => number,
   minCompressRange: number,
-): Record<number, { pending: number; targetBlocks: CompressionBlock[] }> {
-  const out: Record<number, { pending: number; targetBlocks: CompressionBlock[] }> = {};
+): Record<number, TierPending> {
+  const out: Record<number, TierPending> = {};
   const merged = recommendation?.recommendedRanges ?? [];
   const effective =
     minCompressRange > 0
       ? merged.filter((r) => (r.chars ?? r.tokens * 4) >= minCompressRange)
       : merged;
-  out[1] = { pending: effective.reduce((s, r) => s + r.tokens, 0), targetBlocks: [] };
+  out[1] = { pending: effective.reduce((s, r) => s + r.tokens, 0), targetBlocks: [], targetBlockStats: [] };
   const active = activeBlocks(state);
   const t1 = active.filter((b) => b.tier === 1);
   const t2 = active.filter((b) => b.tier === 2);
-  out[2] = { pending: t1.reduce((s, b) => s + countTokens(b.summary), 0), targetBlocks: t1 };
-  out[3] = { pending: t2.reduce((s, b) => s + countTokens(b.summary), 0), targetBlocks: t2 };
+  const t1Stats = t1.map((b) => ({ blockId: b.blockId, summaryTokens: countTokens(b.summary) }));
+  const t2Stats = t2.map((b) => ({ blockId: b.blockId, summaryTokens: countTokens(b.summary) }));
+  out[2] = {
+    pending: t1Stats.reduce((s, st) => s + st.summaryTokens, 0),
+    targetBlocks: t1,
+    targetBlockStats: t1Stats,
+  };
+  out[3] = {
+    pending: t2Stats.reduce((s, st) => s + st.summaryTokens, 0),
+    targetBlocks: t2,
+    targetBlockStats: t2Stats,
+  };
   return out;
 }
 
@@ -1093,6 +1110,7 @@ function decideNudge(input: NudgeInput): NudgeDecision {
     compressibleRanges: rec?.recommendedRanges ?? [],
     protectedRanges: rec?.contextRanges.protected ?? [],
     tierTargetBlocks: injectedTier ? tiers[injectedTier]!.targetBlocks : [],
+    tierTargetBlockStats: injectedTier ? tiers[injectedTier]!.targetBlockStats : [],
     contextUsage: usage,
     tier: injectedTier,
     breakdown: {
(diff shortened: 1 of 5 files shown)


ranxianglei (approved):
**Review: ✅ Approve** — 2 independent agent reviews, both approve. Nits below are optional, none blocking.

**Verified locally on the PR branch** (branch current with master incl. v0.0.31):

- `npm run typecheck` ✅
- `npm test` → 396/396 ✅ (master 391 + 5 new; PR body's "390/390" just reflects older master at authoring time — count is consistent)
- `npm run build` ✅
- CI green (pr-validation, test 22/24); mergeable CLEAN; both `issue #45` named regression tests pass in isolation
- Caller audit: `renderNudgeText`/`formatTierTargetBlocks` have a single render path in src (exported via src/index.ts:35); no other estimator call sites remain

**Design** — implements #45's preferred option exactly:

- Precompute once in the decision layer (`pendingByTier`, src/compress.ts) with the active injected `countTokens`; renderer stays dependency-free and consumes plain data (src/nudge-text.ts:39-46). Displayed number can no longer diverge from the arbitration number.
- `pendingT2`/`pendingT3` now reduce over the same per-block stats (src/compress.ts:961-967) — gate inputs are value-identical to master; no behavior change to arbitration.
- `TierTargetBlockStat` not persisted on `CompressionBlock` — correct, since `countTokens` is host-injected and a stored value would go stale (rationale documented in src/types.ts).
- Renderer joins stats by `blockId` via Map, never by index — ordering drift cannot misattribute numbers. blockIds are unique by construction (monotonic counter, src/state.ts:16-23), so no duplicate-key risk. `??` correctly preserves a legitimate `0` token count.
- Legacy `length / 4` retained only as fallback for hand-built decisions without stats — backward compatible, and covered by a dedicated compat test.

**Tests** — good coverage: custom tokenizer that differs from both length/4 and defaultCountTokens (nice fake-fix guard), CJK-aware default tokenizer end-to-end (decision stats → pendingT2 aggregation → rendered text), legacy fallback, and ASCII-identity when stats equal the legacy estimate.

**Nits (optional)**

1. No tier-3 stat assertion (code is symmetric with tier-2 — low risk).
2. No test for a partial-stats array (one block missing from the Map → falls back per-block).
3. Test-only `!` non-null assertions — pre-existing pattern in this suite, fine as-is.

Fixes #45 as claimed. Thanks!

ranxianglei:
https://github.com/ranxianglei/acp-kernel/pull/91	废弃	远端分支已删无法 rebase；改动(nudge tokens 用 countTokens, https://github.com/ranxianglei/acp-kernel/issues/45)未进 master → 见 issue https://github.com/ranxianglei/acp-kernel/issues/45 需新开小 PR

PR #94, opened 2026-08-21
fix(nudge): wire tiers.*Trigger block-count distillation (T1 treadmill) + snap consumed anchors

ranxianglei:
Closes the structural half of billion-context-pi#3 (repeated-compression reports). Two commits:

## 1. fix(nudge): fire T2/T3 by block count (`14e48ac`)

**Problem — the T1 treadmill.** `tiers.tier2Trigger`/`tier3Trigger` existed in config (defaults 5/10, validated ≥1) but were **never read anywhere** — dead knobs. T2 distillation could only fire on the summary-mass path:

```
t2Pen >= 1.5 × adaptiveGrowth  &&  t2Pen > t1Eff
```

At 1M-window models adaptiveGrowth clamps to growthCap 50K → **75K absolute floor**, plus it must beat T1 effective pending. Production data (27 sessions, billion-context-pi#3): summary mass averages ~235 tok/block; the heaviest session reached **156 active T1 blocks / 36.5K mass — under half the floor** — and 440 blocks across the dataset were **100% Tier-1, zero T2/T3 ever**. Pressure mode (usage ≥ 75%) does arbitrate by max-pending, but 1M sessions peaked at 34% — it never engaged. Summaries accumulate forever → the base keeps growing → nudges keep firing → the model keeps compressing (the reported "连续重复压缩").

**Fix.** Wire the existing knobs into `decideNudge`'s growth branch: T2 ready when active T1 blocks ≥ `tier2Trigger`, T3 when active T2 blocks ≥ `tier3Trigger` — bypassing the mass gates (per-tier cadence still applies, so at most one distill nudge per growthFloor). Mass path unchanged.

**Probe (production parameters, 1M limit, 34% usage, 58 T1 blocks / 13K mass):**
```
tc=363000: inject=true tier=2 "T2 distill ready (block count): 58 tier-1 blocks (mass 13050) >= trigger 5"
tc=365000: inject=false  (cadence: growth 2000 < floor 22500)
```

## 2. fix(boundaries): snap consumed anchors (`79340d3`, was stranded on local master)

Block-boundary distillation misrejected on the pruned view: a block's anchor resolved via `earliestIndexOfIds(effectiveMessageIds)` returns null once its messages are consumed → `not found in visible context (consumed)` (kernel #32). With commit 1 actually suggesting T2, this becomes load-bearing: without it, `compress(b1..b3)` still fails. Snap consumed anchors to the active owning block instead.

## Tests

- 3 new cases in `tests/nudge.test.ts`: T2 by block count below mass threshold; quiet below trigger; T3 by block count.
- Full suite: **391/391 pass**, `tsc --noEmit` clean, build OK.

Merge: human-only per AGENTS.md. After release (0.0.30), billion-context-pi bumps per the kernel-first publish order.

### DIFF
--- a/src/boundaries.ts
+++ b/src/boundaries.ts
@@ -72,6 +72,7 @@ export interface ResolvedRange {
   nestedBlockIds: string[];
   boundaryKind: BoundaryKind;
   protectedGaps: number[];
+  snappedBoundaries: string[];
 }
 
 export function resolveBoundaries(
@@ -90,8 +91,13 @@ export function resolveBoundaries(
     indexByRawId.set(message.id, index),
   );
 
-  let startIndex = resolveAnchorIndex(start, input.state, indexByRawId, "start");
-  let endIndex = resolveAnchorIndex(end, input.state, indexByRawId, "end");
+  let snappedBoundaries: string[] = [];
+  const startAnchor = resolveAnchorIndex(start, input.state, indexByRawId, "start");
+  if (startAnchor.snapped) snappedBoundaries.push(startAnchor.snapped);
+  const endAnchor = resolveAnchorIndex(end, input.state, indexByRawId, "end");
+  if (endAnchor.snapped) snappedBoundaries.push(endAnchor.snapped);
+  let startIndex = startAnchor.index;
+  let endIndex = endAnchor.index;
 
   if (startIndex > endIndex) {
     [startIndex, endIndex] = [endIndex, startIndex];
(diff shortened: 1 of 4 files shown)


ranxianglei:
Closing per @dog's review in billion-context-pi#3: the block-count trigger (`14e48ac`) should not enter master/release — re-analysis of the session dumps confirms the repeated-compression reports were model-initiated (96% of zero-token applies had no nudge within 3 turns; nudges were active on only 1.3% of turns), so wiring `tiers.*Trigger` is unjustified behavior change for now. The commit stays shelved on this branch.

The anchor-snap fix (kernel #32) — a demonstrated bug independent of that debate — is rescoped into #95 (single commit, 388/388 tests). Kernel release 0.0.30 from #95 unblocks the billion-context-pi dependency bump.

PR #100, opened 2026-08-21
fix: validate nudge numeric config fields

### LINKED ISSUE #46: validateConfig() does not sanity-check the newer nudge numeric fields
## Context

v0.0.17, `b9f9e99`. `validateConfig()` (`src/config.ts:44-69`) only validates:

- `modelContextLimit` is finite & positive
- `nudge.minContextLimitPct <= nudge.maxContextLimitPct`
- `promotionThreshold >= 1`
- `truncate.threshold` in `(0, 1]`
- tier trigger ordering (`tier3Trigger > tier2Trigger`, both `>= 1`)

But the nudge config block (`src/types.ts:76-93`) has grown a set of numeric fields that are **never validated**:

- `growthRatio` (default 0.05)
- `growthFloor` (default `max(20000, round(limit*0.05))`)
- `growthCap` (default 50000)
- `minGrowthFloor` (default 20000)
- `minGrowthRatio` (default 0.45)
- `emergencyThresholdPct` (default 0.80)

## Why it matters

`resolveAdaptiveGrowth` (`src/compress.ts:798-810`) is:

```ts
return Math.min(
  nudge.growthCap,
  Math.max(
    nudge.growthFloor,
    Math.round(modelContextLimit * nudge.growthRatio),
  ),
);
```

If a user sets `growthCap < growthFloor` (an easy typo), `Math.min(cap, Math.max(floor, …))` silently returns `growthCap`, which is then below the floor the user intended — the clamp logic masks the misconfiguration instead of flagging it. There is no signal at all; the thresholds just fire at surprising token counts.

Similarly, `emergencyThresholdPct` of `-0.5` or `1.5` is nonsense but accepted: `decideNudge` computes `usage >= config.nudge.emergencyThresholdPct`, so a value `> 1` never fires and a value `<= 0` fires every turn.

`processTurn` only `console.warn`s on validation errors (`src/compress.ts:244-245`) rather than throwing, so even adding checks only makes the problem *visible* — it does not hard-fail. That is the current design choice and fine; the gap is that the visibility does not exist for these fields at all.

## Suggested checks to add

```ts
if (config.nudge.growthFloor < 0) errors.push("nudge.growthFloor must be >= 0");
if (config.nudge.growthCap < config.nudge.growthFloor)
  errors.push("nudge.growthCap must be >= nudge.growthFloor");
if (config.nudge.growthRatio < 0)
  errors.push("nudge.growthRatio must be >= 0");
if (config.nudge.minGrowthFloor < 0)
  errors.push("nudge.minGrowthFloor must be >= 0");
if (config.nudge.minGrowthRatio < 0)
  errors.push("nudge.minGrowthRatio must be >= 0");
if (config.nudge.emergencyThresholdPct <= 0 || config.nudge.emergencyThresholdPct > 1)
  errors.push("nudge.emergencyThresholdPct must be in (0, 1]");
```

## Severity

Low. Affects only users who override defaults (the defaults are all internally consistent). But config validation is cheap defense-in-depth, and the `growthCap < growthFloor` case in particular is silently masked rather than noisy — exactly the kind of thing a validator should catch.

Beatrice0377:
## Problem

`validateConfig()` does not sanity-check several numeric nudge configuration fields added over time.

Invalid overrides such as a negative growth floor, `growthCap < growthFloor`, or an out-of-range emergency threshold are currently accepted and can make nudge thresholds behave unexpectedly without any validation warning.

Non-finite values such as `NaN` or `Infinity` can also bypass ordinary numeric comparisons and propagate into the adaptive-growth calculations.

## Root Cause

The newer nudge numeric fields were added without corresponding checks in `validateConfig()`.

The existing validator covers model context limits, tier trigger ordering, truncation thresholds, and some nudge threshold relationships, but not:

- `growthRatio`
- `growthFloor`
- `growthCap`
- `minGrowthFloor`
- `minGrowthRatio`
- `emergencyThresholdPct`

## Fix

Add validation for the six fields above:

- Require all six values to be finite numbers.
- Require `growthRatio`, `growthFloor`, `minGrowthFloor`, and `minGrowthRatio` to be non-negative.
- Require `growthCap >= growthFloor`.
- Require `emergencyThresholdPct` to be in `(0, 1]`.

The validator keeps its existing behavior of returning validation errors; no runtime path is changed to throw.

## Scope / Non-goals

- No changes to nudge runtime or arbitration behavior.
- No changes to default configuration values.
- No validation policy is added for `tier2GrowthMultiplier` or other fields outside the scope of #46.
- No version, dependency, lockfile, workflow, or documentation changes.

## Tests

Added regression coverage for:

- Negative nudge growth values.
- `NaN` / `Infinity` inputs.
- `growthCap < growthFloor`.
- `growthCap == growthFloor`.
- Zero-valued non-negative fields.
- `emergencyThresholdPct` boundaries.
- Compatibility for finite growth ratios above `1`.

Validation:

- Targeted config tests: 18/18 passed.
- Full test suite: 400/400 passed.
- `npm run typecheck`: passed.
- `npm run build`: passed.
- `git diff --check`: passed.

## Compatibility

Valid existing configurations continue to pass validation.

The change only adds warnings for invalid configuration values; it does not alter nudge calculation behavior for valid inputs.

Fixes #46

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -1,4 +1,4 @@
-import type { Config } from "./types.js";
+import type { Config, NudgeConfig } from "./types.js";
 
 export function defaultConfig(
   modelContextLimit: number,
@@ -71,6 +71,34 @@ export function validateConfig(config: Config): string[] {
   }
   if (config.tiers.tier3Trigger <= config.tiers.tier2Trigger) {
     errors.push("tiers.tier3Trigger must be greater than tiers.tier2Trigger");
+  }
+  const nonNegative: Array<
+    keyof Pick<
+      NudgeConfig,
+      "growthRatio" | "growthFloor" | "minGrowthFloor" | "minGrowthRatio"
+    >
+  > = ["growthRatio", "growthFloor", "minGrowthFloor", "minGrowthRatio"];
+  for (const field of nonNegative) {
+    const v = config.nudge[field];
+    if (!Number.isFinite(v)) {
+      errors.push(`nudge.${field} must be a finite number`);
+    } else if (v < 0) {
+      errors.push(`nudge.${field} must be >= 0`);
+    }
+  }
+  if (!Number.isFinite(config.nudge.growthCap)) {
+    errors.push("nudge.growthCap must be a finite number");
+  } else if (
+    Number.isFinite(config.nudge.growthFloor) &&
+    config.nudge.growthCap < config.nudge.growthFloor
+  ) {
+    errors.push("nudge.growthCap must be >= nudge.growthFloor");
+  }
+  const emergency = config.nudge.emergencyThresholdPct;
+  if (!Number.isFinite(emergency)) {
+    errors.push("nudge.emergencyThresholdPct must be a finite number");
+  } else if (emergency <= 0 || emergency > 1) {
+    errors.push("nudge.emergencyThresholdPct must be in (0, 1]");
   }
   return errors;
 }
--- a/tests/sync-config.test.ts
+++ b/tests/sync-config.test.ts
@@ -93,3 +93,111 @@ test("validateConfig flags min > max nudge thresholds", () => {
 test("validateConfig passes for default config", () => {
   assert.deepEqual(validateConfig(defaultConfig(200000)), []);
 });
+
+// Issue #46 scope: tier2GrowthMultiplier is intentionally left unvalidated.
+type NudgeNumberField =
+  | "growthRatio"
+  | "growthFloor"
+  | "growthCap"
+  | "minGrowthFloor"
+  | "minGrowthRatio"
+  | "emergencyThresholdPct";
+
+test("validateConfig rejects negative nudge growth values", () => {
+  const cases: Array<{ field: NudgeNumberField; value: number }> = [
+    { field: "growthRatio", value: -0.1 },
+    { field: "growthFloor", value: -1 },
+    { field: "minGrowthFloor", value: -1 },
+    { field: "minGrowthRatio", value: -0.1 },
+  ];
+  for (const { field, value } of cases) {
+    const cfg = defaultConfig(200000);
+    cfg.nudge[field] = value;
+    assert.ok(
+      validateConfig(cfg).some((e) => e.includes(`nudge.${field}`)),
+      `expected nudge.${field} to be rejected`,
+    );
+  }
+});
+
+test("validateConfig rejects non-finite nudge numbers", () => {
+  const cases: Array<{ field: NudgeNumberField; value: number }> = [
+    { field: "growthRatio", value: NaN },
+    { field: "growthFloor", value: Infinity },
+    { field: "growthCap", value: Infinity },
+    { field: "minGrowthFloor", value: NaN },
+    { field: "minGrowthRatio", value: Infinity },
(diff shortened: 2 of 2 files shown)


ranxianglei (approved):
**Review: ✅ Approve** — 2 independent agent reviews, both approve. Nits below are optional follow-ups, none blocking.

**Verified locally on the PR branch** (claims all check out):

- `npm run typecheck` ✅
- `npm test` → 400/400 ✅ (matches PR claim)
- `npm run build` ✅
- CI green (pr-validation, test 22/24); mergeable; single commit; scope confirmed: `src/config.ts` (+29/−1), `tests/sync-config.test.ts` (+108)

**Correctness**

- Implements all six checks from #46, and adds finiteness (NaN/Infinity) checks beyond the issue's suggestion — good hardening, since NaN silently bypasses `<`/`>` comparisons.
- The `Number.isFinite(growthFloor)` guard at src/config.ts:92 avoids a confusing "growthCap must be >= growthFloor" cascade when the floor itself is non-finite. Probed: `growthFloor: Infinity` → single clear error; `growthFloor: NaN, growthCap: -1` → root-cause error only.
- `growthCap >= 0` isn't checked directly but holds transitively (cap ≥ floor ≥ 0). Negative cap with valid floor is caught via the cap-vs-floor relation.
- `emergencyThresholdPct` domain `(0, 1]` is correct: `usage >= emergencyThresholdPct` (src/compress.ts:967) means 0 would fire every turn and >1 never fires — exactly the nonsense values #46 wanted flagged.
- Semantics vs `resolveAdaptiveGrowth` (src/compress.ts:922-928): with cap < floor, `Math.min(cap, Math.max(floor, …))` silently returns cap — the validator now surfaces that misconfiguration instead of masking it.
- Defaults (growthFloor=50000, growthCap=50000, emergencyThresholdPct=0.95) validate clean; also probed `defaultConfig(2_000_000)` to rule out limit-dependent breakage.
- No consumer risk: `validateConfig` output is only `console.warn`'d (src/compress.ts:327-329), never thrown; pure additive checks, no behavior change for valid configs.

**Nits (optional)**

1. No explicit negative-`growthCap` test (`growthCap: -1` with a valid floor) — tests only cover cap=99/floor=100.
2. `growthCap: NaN` untested (only `Infinity`, same `!Number.isFinite` branch — cosmetic).
3. `emergencyThresholdPct: -0.5` untested (0 and 1.01 already hit both branches — cosmetic).
4. Follow-up candidate (explicitly out of scope here): `tier2GrowthMultiplier`, `frequency`, `iterationThreshold`, and min/max pct ranges remain unvalidated.

The `keyof Pick<NudgeConfig, …>` typing (src/config.ts:75-80) is a nice touch — keeps the field list refactor-safe. Thanks!

ranxianglei:
Closing as superseded: master `src/config.ts:63-75` already validates the nudge numeric fields — ordering (`minContextLimitPct <= maxContextLimitPct <= emergencyThresholdPct`) and `Number.isFinite` checks (e.g. `minPressureBenefitTokens`). Thanks for the contribution — the validation surface you proposed is in place. If you spot a field that still slips through unvalidated, please open a fresh PR against master.

PR #103, opened 2026-08-21
compress: break identical-rejection retry death-loops

ranxianglei:
## Incident

pi session `01a02542` (qwen3.8-27b via local vLLM/SGLang, billion-context-pi 0.1.43 + acp-kernel 0.0.30) entered a compress retry death-loop:

- After 5 successful compressions nothing compressible remained (22.6K context, 11 active blocks, remaining compressible content 0 chars < min 5000).
- The min-gate rejection ended with: *"Nothing to do. Current active blocks span b1..b11 — **retry** with startId/endId set to active block IDs in that span."*
- The model re-submitted the **identical** rejected call () **3853 times** over **5h11m** (~4.9s/call, ~800 GPU-minutes burned) — it followed the literal "retry" instruction and ignored the "Nothing to do" half. No dedup or circuit breaker existed anywhere.

## Fix

1. **Rejection tracking** — `CompressionState.rejections` (new optional field; bounded 8-entry FIFO of `{ spec, count }`; keyed on the range signature startRef..endRef list; backward compatible — old persisted states start counting from 1, cloned by `cloneState`).
2. **Terminal message from the 2nd identical rejection** — the gate responds: identical range(s) rejected N times, re-submitting will keep failing, stop calling compress with these refs and answer the user; new content becomes compressible once it exceeds min chars. **No retry guidance of any kind.**
3. **First-rejection advice kept, de-weaponized** — the liveHint now reads "only a block-to-block distillation (startId/endId set to those block IDs) can reclaim more; otherwise nothing remains to compress" — the word "retry" is gone, the genuinely useful tier-2 action is still offered.

## Tests

- New `tests/compress-rejection-loop.test.ts` (4 cases): first rejection keeps guidance + records count; identical 2nd/3rd rejections flip to terminal (asserts the message contains "rejected N times" and contains NO retry/re-issue/acp_status guidance); different range gets its own counter; caller state not mutated; legacy state without the field counts from 1.
- Updated the one assertion that pinned the old "retry with startId/endId" wording.

Full pre-flight: `tsc --noEmit` clean, **395 tests pass**, build OK.

## Blast radius

Pure kernel gate change — both adapters (billion-context-pi, billion-context-omp) pick it up on their next kernel pin bump. The omp adapter's compress tool needs no change (it forwards kernel errors verbatim).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "acp-kernel",
-    "version": "0.0.28",
+    "version": "0.0.31",
     "lockfileVersion": 3,
     "requires": true,
     "packages": {
         "": {
             "name": "acp-kernel",
-            "version": "0.0.28",
+            "version": "0.0.31",
             "license": "MIT",
             "devDependencies": {
                 "@types/node": "^22.0.0",
--- a/src/compress.ts
+++ b/src/compress.ts
@@ -246,21 +246,30 @@ export function createCore(ports: Ports = {}): CompressionCore {
         }
       }
       if (!hasBlockBoundaryRange && totalRangeChars < input.config.compress.minCompressRange) {
+        const rejectionSpec = input.ranges.map((r) => `${r.startRef}..${r.endRef}`).join(",");
+        const priorCount = state.rejections?.find((r) => r.spec === rejectionSpec)?.count ?? 0;
+        const rejectionCount = priorCount + 1;
+        state.rejections = [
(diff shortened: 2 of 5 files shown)


ranxianglei:
Closing as superseded: master `src/compress.ts:844` already carries the identical-call death-loop guard ("repeats the identical call forever" — promote/merge must go through the review path). The same failure class is also covered host-side in billion-context (#156 → PR#370 `failedSignatures`).

PR #123, opened 2026-08-22
feat(report): drilldown reverse + offset options — let users reach the newest messages

### LINKED ISSUE #93: enhancement: acp_status drilldown cannot reach the newest messages — sort:time is ascending-only, slice(0,limit) has no offset/reverse
## Problem

`acp_status` drilldown (`scope:"uncompressed"` + `view:"messages"`) with `sort:"time"` always starts from the OLDEST message. With the default `limit: 30` (and any limit < total message count), the NEWEST messages are unreachable — there is no `offset`, `reverse`, or descending-time sort option.

## Mechanism (dist/index.js 0.0.29)

```js
// renderMessageDrilldown
if (sort === "time") filtered.sort((a, b) => a.index - b.index); // ascending only
...
const shown = filtered.slice(0, limit); // head-only, no pagination
```

`collectVisible` assigns `index` = array position (0-based, ascending), so `sort:"time"` shows oldest-first and `slice(0, limit)` truncates the tail. Any limit < total count hides the newest messages.

## Why it matters

- The model cannot locate the newest (often largest, just-executed) tool outputs without a message-stream ref tag. 
- In billion-context-pi the model sees per-message `<acp>` ref tags injected into the message stream (`src/messages.ts patchRefTag`), so drilldown is only an auxiliary view there. Hosts WITHOUT such tags (e.g. billion-context-dsh, decision: no `<acp>` tags, seq-is-the-ref) rely on drilldown as the ONLY message-location entry — the head-only behavior is amplified into a real gap.
- Same limitation applies to `sort:"size"`/`sort:"tool"` (head `limit` rows of the sorted order) — large tool outputs from the tail of the conversation cannot be seen unless the model knows the tool name and uses `tool:"X"` + `sort:"size"`.

## Requested change

Add a way to reach the tail of the list, any of:
1. `reverse: true` option (flip the sort order after sorting)
2. `offset: number` option (pagination)
3. `sort:"time-desc"` (or make `sort:"time"` descending and add `"time-asc"`)

Minimal and backwards-compatible; benefits both pi and DSH hosts.

## Context

Tracking issue in our host repo: https://github.com/Tyan66666/billion-context-dsh/issues/46

Tyan66666:
Closes #93

## Problem

When agent opens the `acp_status` drilldown (`scope:"uncompressed"` + `view:"messages"`, `sort:"time"`), the list always starts at the **oldest** message. `slice(0, limit)` only ever shows the head, so once the conversation is longer than `limit`, the **newest messages are simply unreachable** — there is no way to flip the order (`reverse`), page forward (`offset`), or sort newest-first. The same head-only limitation applies to `sort:"size"` / `sort:"tool"` and to `scope:"compressed"`.

Why it matters differs by host:

- Hosts that inject `<acp>` ref tags into the message stream (e.g. billion-context-pi) see every message inline — the drilldown is a nice-to-have there.
- Hosts without tags (billion-context-dsh: no tags, seq-is-the-ref) rely on the drilldown as their **only** way to locate messages — for them the head-only behavior is a real gap (tracking: Tyan66666/billion-context-dsh#46).

## Fix

Two new optional options on `StatusReportOptions` (backwards-compatible, no behavior change when omitted):

- `reverse?: boolean` — after sorting, flip the order. `sort:"time" + reverse:true` gives newest-first, so the tail of a long conversation becomes reachable.
- `offset?: number` — pagination. Rows are windowed with `slice(offset, offset + limit)`.

Applied in `renderMessageDrilldown` and `renderCompressedDrilldown` (both honor `reverse` then window with `offset`). Footer text: `offset` 0 keeps the existing `N of M shown.`; `offset > 0` prints `Showing A–B of M.`; an offset past the end prints `Offset N past end (M total).` instead of a silent empty page. The overview Tip now advertises `sort:"time" + reverse:true` for newest-first, and README's drilldown options table + example are updated.

Usage:

```ts
buildStatusReport(state, messages, countTokens, {
  scope: "uncompressed",
  view: "messages",
  sort: "time",
  reverse: true,   // newest message first — reach the tail
  limit: 30,
});
// pagination: offset: 30 → the next page
```

## Tests

7 new cases in `tests/report-drilldown.test.ts`: status-quo pin (oldest-first head-only), `reverse:time` reaches the newest messages, offset pagination in both directions, offset past end, `sort:"size"` reverse (smallest-first), `tool` filter + reverse, and `scope:"compressed"` reverse + pagination. Full suite: 444 pass, typecheck + build clean.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ tmp/
 .DS_Store
 coverage/
 .acp-*.json
+worktrees/
--- a/README.md
+++ b/README.md
@@ -95,17 +95,15 @@ live-recomputed tags, use `renderVisibleRefs` directly.
 
 #### `buildStatusReport` drilldown — reaching the newest rows
 
-`buildStatusReport` accepts optional `scope`/`view`/`tool`/`sort`/`limit`
-plus two navigation options:
-
-| Option | Default | Effect |
-|--------|---------|--------|
-| `reverse: boolean` | `false` | Flip the sorted order after sorting — `sort:"time"` + `reverse:true` lists newest-first, `sort:"size"` + `reverse:true` lists smallest-first |
-| `offset: number` | `0` | Skip the first N rows of the sorted order (pagination with `limit`) |
+The per-message drilldown (`scope:"uncompressed"` + `view:"messages"`) sorts
+`time` ascending by default and shows only the first `limit` rows, so the
+newest messages are unreachable once the conversation outgrows the limit.
+Two optional navigation options fix that: `reverse` (default `false`) flips
+the sorted order after sorting — `sort:"time"` + `reverse:true` lists
+newest-first, `sort:"size"` + `reverse:true` lists smallest-first — and
+`offset` (default `0`) skips the first N rows, paginating with `limit`.
 
 ```ts
-import { buildStatusReport } from "acp-kernel";
-
 buildStatusReport(state, messages, countTokens, {
   scope: "uncompressed",
   view: "messages",
@@ -116,11 +114,10 @@ buildStatusReport(state, messages, countTokens, {
 // pagination: offset: 30 → the next page
 ```
 
-Sort orders are otherwise unchanged (`time` ascending, `size` largest-first,
-`tool` grouped, `age` most-survived-first for `scope:"compressed"`), so existing
-callers keep working. The "N of M shown." footer becomes
-"Showing A–B of M." when `offset > 0`, and reports "Offset N past end" when the
-window starts beyond the list — no silent empty page.
+`reverse` and `offset` are optional, so existing call sites keep working
+unchanged. The "N of M shown." footer stays for `offset: 0`; when
+`offset > 0` it reads "Showing A–B of M.", and an offset past the end reports
+"Offset N past end" instead of a silent empty page.
 
 ### Standalone modules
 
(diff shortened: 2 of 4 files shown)


ranxianglei:
Review note (non-blocking) — from the ework review thread (dog/acp-kernel#2):

The new tip line in `renderOverview` (src/report.ts:205) now tells the agent reading the status report to call with `sort:"time", reverse:true`. However, host adapters' status tools may not yet expose `reverse`/`offset` as parameters — e.g., the ework daemon's `acp_status` tool currently only exposes `scope`/`view`/`tool`/`sort`/`limit`. Until adapters plumb the new options through, the tip will recommend something the model cannot actually pass.

Kernel-side behavior is correct and backward-compatible, so this does not block merging — but please either:
1. note the cross-repo dependency in the PR description, and/or
2. file a follow-up in the adapter repo(s) (e.g., pai-acp) to expose `reverse`/`offset` on their status tool surface.

ranxianglei:
https://github.com/ranxianglei/acp-kernel/pull/123	废弃	远端分支已删 且 CI 失败(pr-validation/build-artifact)；改动(drilldown reverse/offset, https://github.com/ranxianglei/acp-kernel/issues/93)未进 master → 见 issue https://github.com/ranxianglei/acp-kernel/issues/93

PR #130, opened 2026-08-22
feat(persist): legacy record adoption + loadSync path hint

ranxianglei:
Extends the `acp-kernel/persist` StateStore (0.0.40) so downstreams owning a pre-envelope on-disk format can adopt it:

- `legacy` option: called for any file that fails the envelope-shape check; return `{id, payload, version?, savedAt?}` to load it as an envelope. Source stamps are preserved; the next dirty write re-persists as an envelope — files migrate organically, old files keep loading.
- `loadSync(id, hint)`: probe an extra relative path for namespaced records the store has not discovered (a store that did not run loadAll at boot, e.g. an evicted session re-requested with its meta).

Unblocks billion-context proxy's SessionStore swap to the kernel store (its on-disk format is a flat v3 record). 463/463.

### DIFF
--- a/src/persist/store.ts
+++ b/src/persist/store.ts
@@ -18,6 +18,15 @@ export interface PersistedEnvelope<T> {
     payload: T;
 }
 
+/** A legacy (pre-envelope) record adopted on load. `version`/`savedAt`
+ * preserve the source record's own stamps when present. */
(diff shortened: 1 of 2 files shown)


ranxianglei:
这个是模型越权合并,可能存在风险.越权模型glm5.3

PR #137, opened 2026-08-23
fix(compress): clarify stale-ref error to point at acp_status

ranxianglei:
**Model: qwen3.8-27b (vllm)**

## Problem
When a compress call targets a range that is already compressed (stale refs from before a renumbering compress), the case-2 gate message was misleading:

> Requested range(s) already compressed (e.g. m00365..m00462); remaining compressible content 0 chars < min 5000. Nothing to do.

The "remaining compressible content 0 chars < min 5000. Nothing to do" framing reads as *there is nothing to compress* (give up), not *my refs are stale — re-check*. The model then spins on repeated failed compress attempts, each one adding its (unapplied) summary + error to the context as a net token loss. Observed live in a pi session: two failed compresses (m00365..m00462, m00210..m00211) both hit this message, and the model concluded "there's genuinely nothing to compress" instead of re-checking refs.

The all-unknown case (case 1) already has the correct actionable wording; case 2 did not.

## Fix
Rewrite the case-2 gate message (src/compress.ts) to lead with the actionable guidance, matching case 1:

> Requested range(s) already compressed (e.g. m00365..m00462) — your refs are stale: a prior compress renumbered the remaining refs, so this range now falls inside an active block. Run acp_status and re-issue the compress in the same turn using only the CURRENT compressible ranges it reports.

## Tests
- Updated tests/compress.test.ts assertion to verify the new actionable wording ("your refs are stale" + "Run acp_status").
- typecheck clean; all 479 tests pass.

Fixes billion-context-pi#178.

### DIFF
--- a/src/compress.ts
+++ b/src/compress.ts
@@ -272,7 +272,7 @@ export function createCore(ports: Ports = {}): CompressionCore {
           unknownCount > 0
             ? `None of the ${input.ranges.length} requested range(s) resolved — every ref failed with "does not exist in this session". Refs recorded before an earlier compress are stale: each successful compress renumbers the remaining refs. Run acp_status, then re-issue the compress in the same turn using only the refs it reports.`
             : consumedRanges.length > 0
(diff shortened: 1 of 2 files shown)


ranxianglei:
改动存疑

PR #159, opened 2026-08-23
fix: hard per-tool-result token cap (billion-context-pi incident 2026-08-23)

ranxianglei:
# fix: hard per-tool-result token cap (billion-context-pi incident 2026-08-23)

> Author: **GLM-5.3** (zhipuai-lb), via pi coding agent — per repo AGENTS.md rule 2.
> Base: `2026-08-23_truncate-size-aware` (stacks on open PR #133; all of its changes kept intact).

## Problem

Real incident (2026-08-23): a pi session with ACP died. A single bash toolResult of
50,358 chars of minified JS (~31,475 tokens by ACP's own count, ~24% of the effective
131,072-token window) entered context **whole**:

- Byte-based limits never fired — it was under pi's 51,200-byte bash cap and under ACP's
  200,000-byte cap.
- Token-percentage valves never fired — ACP's estimate read **51.8%** usage while real
  usage was **>100%**, so the 95% emergency truncation (and every nudge threshold) stayed
  closed. The model config was `contextWindow=262144/maxTokens=131072` on sglang enforcing
  `input + max_tokens ≤ 262,144`; the server then returned `400 status code (no body)`
  forever — a dead loop.

A single message can consume a double-digit share of the window while *total* usage still
reads low. No usage-gated valve — emergency truncation included — can catch that class of
failure. It needs a hard **per-message** guard.

## Fix

New `tool-result-cap` pipeline node + config knob:

- **Config**: `truncate.maxToolResultTokens`
  - `null` (default) = auto: `min(10% × modelContextLimit, 16384)` tokens
  - `0` = disabled
  - explicit positive number = that cap
  - validated in `validateConfig` (negative / non-finite → error)
- **`capLargeToolResults`** (`src/truncate-tools.ts`): any tool-result whose estimated
  tokens exceed the cap is rewritten head+tail with the marker line
  `[acp: tool-result truncated, original ~N tokens]` (reuses the existing truncation
  text style). Sizing starts from a ~4 chars/token budget and halves until the
  CJK-aware count fits, so CJK-dense results (1 char ≈ 1 token) are capped correctly.
- **Node placement**: after `prune` (no point rewriting messages about to disappear) and
  before `recommend`/`nudge-inject` (downstream estimates reflect the capped sent view).
  Runs on **every** turn — no usage gate — and **ignores recency protection**: the cap is
  a hard guarantee about the outgoing context, not a suggestion.
- Estimation uses the same CJK-aware tokenizer as the rest of the kernel
  (`ports.countTokens ?? defaultCountTokens`) — a naive `chars/4` estimator would pass a
  30,000-CJK-char result that is really 30,000 tokens.
- Emergency truncation now **skips already-capped texts** (marker check), so the two
  valves never stack markers or double-truncate.
- Applies to ALL tool-results, including protected tools — the 2026-08-22 overflow was
  caused precisely by large *recent* decompress inline results.

PR #133's size-aware emergency truncation is untouched apart from the extra skip
condition in its candidate scan.

## Tests (`tests/toolresult-token-cap.test.ts`, 13 new + 1 in pipeline.test.ts)

- auto cap math incl. the 10% rule and the 16384 ceiling; unknown limit falls back to the ceiling
- explicit cap, disable (`0`), explicit-overrides-auto
- oversized fresh tool-result rewritten; head+tail survive; result fits cap under the
  CJK-aware tokenizer; small ones untouched (same array reference)
- CJK content caught where `chars/4` would pass it
- `processTurn` integration: caps the MOST RECENT tool-result at 20% usage — far below
  every usage-gated valve — with no emergency-truncation marker
- idempotency (already-capped text is skipped)
- `truncateLargeToolOutputs` skips capped texts (no double truncation)
- `validateConfig` rejects negative values; `null`/`0` valid

Full suite: **477 pass / 0 fail**; `tsc --noEmit` clean; prettier clean on touched files.

## Notes / deviations

- The `<acp>` tag render-refs prepends after the cap node is not counted against the cap
  — consistent with the kernel's own accounting (`renderMessage` strips the own tag
  before counting).
- Marker wording matches the requested `[acp: tool-result truncated, original ~N tokens]`,
  wrapped in the existing `...`-style truncation decoration.

### DIFF
--- a/src/compress.ts
+++ b/src/compress.ts
@@ -376,6 +376,7 @@ export function createCore(ports: Ports = {}): CompressionCore {
       messages: result.messages,
       state: result.state,
       nudge: result.effects.nudge,
+      toolResultCappedCount: result.effects.toolResultCappedCount,
     };
   }
 
--- a/src/compress.ts
+++ b/src/compress.ts
@@ -11,7 +11,10 @@ import {
   blockVisibleInRange,
 } from "./boundaries.js";
 import type { ResolvedRange } from "./boundaries.js";
-import { truncateLargeToolOutputs } from "./truncate-tools.js";
+import {
+  truncateLargeToolOutputs,
+  capLargeToolResults,
+} from "./truncate-tools.js";
 import { hideConsumedCompressCalls } from "./hide-consumed.js";
 import { applyMessageFilters, listMessageFilters } from "./filter/index.js";
 import { createRenderRefsNode } from "./render-refs.js";
@@ -426,6 +429,7 @@ export function createCore(ports: Ports = {}): CompressionCore {
       pruneNode,
       filterNode,
       hideCompressCallsNode,
+      toolResultCapNode,
       recommendNode,
       nudgeNode,
       emergencyTruncateNode,
@@ -585,6 +589,29 @@ const nudgeNode: PipelineNode = {
       ...io,
       state: { ...io.state, nudge: stamped },
       effects: { ...io.effects, nudge },
+    };
+  },
+};
+
+// Hard per-tool-result token cap. Runs on EVERY turn (no usage gate) and
+// ignores recency protection: a single oversized tool-result can consume a
+// double-digit share of the window while total usage still reads low, so no
+// percentage-gated valve can catch it. Placed after prune (no point rewriting
+// messages about to disappear) and before recommend/nudge (so downstream
+// token estimates reflect the capped sent view).
+const toolResultCapNode: PipelineNode = {
+  name: "tool-result-cap",
+  run(io, ctx) {
+    const capped = capLargeToolResults(
+      io.messages,
+      ctx.config,
+      ctx.countTokens,
+    );
+    if (capped.cappedCount === 0) return io;
+    return {
+      ...io,
+      messages: capped.messages,
+      effects: { ...io.effects, toolResultCappedCount: capped.cappedCount },
     };
   },
 };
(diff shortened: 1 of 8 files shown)


ranxianglei:
Review round 1 (agent reviewer + fixes applied in `ffc541d`):

**Cache stability verdict: the design holds** — deterministic stateless projection, token-tag snapshot frozen, tokenizer bound once per core. Two real bugs + one cache hazard found and fixed:

1. **Marker-bypass (major)**: a legitimate tool-result *quoting* `[acp: tool-result truncated` (grep of this repo, git diff, log dump) escaped the cap entirely. Now: marker-marked text is skipped **only while it still fits** (≤ cap × 1.25 drift margin).
2. **NaN cap (major)**: non-finite `maxToolResultTokens`/`modelContextLimit` compared false against every bound → *every* tool-result replaced by the bare marker (`hello world` included). Falls back to auto now.
3. **Cap movement (major)**: auto cap moved with the learned context limit → prefix break per limit change, per-turn for hosts that re-derive the limit per request. Auto cap quantized down to powers of two (step 1024): truncation point moves only across power-of-two boundaries.
4. **Emergency re-truncation**: cap-marked messages are no longer exempt from ≥95% emergency truncation (they were permanently untouchable before).
5. `toolResultCappedCount` surfaces on `ProcessTurnResult` for host observability; cheap char prefilter (4 tokens/char ceiling) skips re-tokenizing small results.

482/482 tests pass (5 new regression tests).

PR #165, opened 2026-08-26
fix(report): group acp_status uncompressed ranges by conversation turn

ranxianglei:
## Problem

`acp_status({scope:"uncompressed"})` (the default ranges view) collapsed the entire visible window into ONE giant range whenever message refs were dense, because `renderUncompressedRanges` only merged consecutive ref numbers and never split at conversation boundaries. The output stopped being actionable, and models fell back to `view:"messages"` (the per-message firehose) - see ranxianglei/billion-context-pi#193.

## Fix (src/report.ts)

- Split ranges at user-message boundaries once a group has >= 3 messages - the same condition `buildCompressibleRanges` already uses (recommend.ts), so each range aligns to roughly one conversation turn.
- Sort ranges by size by default (largest first, tie-broken by start ref) so the view directly answers "what should I compress first"; `sort:"time"` keeps chronological order. A `Sorted by size|time` header makes the ordering explicit.
- Honor the `limit` option instead of a hardcoded 30.

## Tests

4 new tests in tests/decompress-report.test.ts: turn-boundary splitting, size sort + tie-break, sort=time order, limit truncation.

Rebased onto master (v0.0.74); conflicts resolved in favor of the newer dominant-tool (`toolTokens`) tracking while keeping the turn-boundary split. Full suite: 793 pass, 0 fail; typecheck + build clean.

Note: repo-wide format:check fails on many files on master (pre-existing; files use 4-space indent vs prettier default 2). New code matches the file's existing style.

### DIFF
--- a/src/report.ts
+++ b/src/report.ts
@@ -61,6 +61,7 @@ interface VisibleMessageInfo {
     tokens: number;
     tool: string;
     index: number;
+    isUser: boolean;
 }
 
 function collectVisible(
@@ -84,7 +85,7 @@ function collectVisible(
         if (!ref) return;
         const tokens = countTokens(message.text ?? "");
         const tool = message.toolName ?? "text";
-        if (tokens > 0) visible.push({ ref, tokens, tool, index });
+        if (tokens > 0) visible.push({ ref, tokens, tool, index, isUser: message.role === "user" });
     });
     return { visible, summaryTokens };
 }
@@ -123,7 +124,7 @@ export function buildStatusReport(
         if (view === "messages") {
             return renderMessageDrilldown(visible, toolFilter, sort, limit);
         }
-        return renderUncompressedRanges(visible);
+        return renderUncompressedRanges(visible, sort, limit);
     }
 
     return renderOverview(visible, summaryTokens, activeBlocks, state, countTokens, limit);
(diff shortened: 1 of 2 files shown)


ranxianglei:
[bot] Reviewed branch `2026-08-26_status-ranges-by-turn` (7fecd7d) against master.

**Verified locally** (fresh clone, `npm ci` + real Node):
- `npm run typecheck` — clean
- `npm test` — **544 pass, 0 fail** (matches the issue's claim)
- `npm run build` — clean

**Code review:**
- Turn-boundary condition in `src/report.ts:229` (`m.isUser && last.count >= 3`) is exactly the condition `buildCompressibleRanges` uses at `src/recommend.ts:216` (`info.isUser && cur.count >= 3`) — the "same condition" claim checks out. Split behavior is also equivalent: the user message starts the new group in both.
- Sort: default size-desc with start-ref tie-break (`src/report.ts:241`), `sort:"time"` keeps chronological order, `Sorted by size|time` header added. One minor note: `StatusReportOptions.sort` also allows `"tool" | "age"`, which fall through to size sort in this view — the header honestly reports "size" in that case, so it's not misleading, just worth knowing.
- `limit` is now honored (`src/report.ts:244`) instead of the hardcoded 30; default stays 30 (`src/report.ts:111`).
- Tests: pure additions in `tests/decompress-report.test.ts` — 4 new tests, no pre-existing assertions modified. Token math in the fixtures checked out (turns = 9/9/23 tokens, so the size-sort and tie-break assertions are meaningful, not accidental).

**Reviewer 1: LGTM.** Per AGENTS.md this needs a second agent review before merge, and merging is human-only — I won't merge it.

PR #171, opened 2026-08-30
fix(prune): keep rendered summaries in the leading system prefix

### LINKED ISSUE #170: prune: summary anchor can land mid-conversation (role:system after a user message) — 400 on strict OpenAI backends
## What

`collectSummaryAnchors` (src/prune.ts, dist/index.js:136-165) sets `insertAt = earliest ?? 0` where `earliest` is the index of the block's earliest covered message **in the input messages**. `rebuildMessages` (dist/index.js:166-189) then inserts the rendered summary — `renderSummary` produces `role: "system"` (dist/index.js:190-201) — at that index.

Whenever the covered range does NOT start at the conversation head, the anchor sits after one or more still-visible messages, and the rebuilt message list carries a **system message mid-conversation**:

```
[user(m00001), assistant(m00002), system(summary), user(m00011), ...]
```

This is not limited to multi-segment compress: **any** compress whose range starts after the head (the normal "fold the old stuff, keep the tail" pattern) produces it. The only safe shape is a range starting at index 0.

Note the kernel already preserves the first user message unconditionally (`rebuildMessages`, dist/index.js:176-179: `if (index === firstUserIndex && firstUserIndex >= 0) { result.push(...); continue; }`), so even a head-anchored range m00001–mNNN leaves m00001 visible — the summary lands at index 0, ahead of it, which is fine. The problem is exclusively anchors at index > 0.

## Impact

- OpenAI chat-wire hosts: `coreToOpenai` serializes `role: "system"` verbatim at its position, so strict OpenAI-compatible backends reject every subsequent request. Reproduced in production with sglang (qwen3.8-27B): `400 System message must be at the beginning.` on ALL requests after the compress; the session is stuck until the provider is switched or the proxy bypassed. See ranxianglei/billion-context#355.
- In-process hosts (pi/opencode plugins) that consume core messages directly carry the mid-conversation system message into whatever wire they build.

## Suggested fix

Clamp the anchor to the first user message in `collectSummaryAnchors` — `firstUserIndex` is already computed in `prune()` (dist/index.js:117-119) but only passed to `rebuildMessages`:

```ts
insertAt: Math.min(earliest ?? 0, firstUserIndex)
```

This folds every summary into the leading system prefix. Semantics are unchanged: a summary is a stand-in for the folded history, and a leading system prefix is accepted by both OpenAI and Anthropic (the anthropic wire already maps non-assistant roles to user, so it is unaffected today).

Alternative considered: render summaries as `role: "user"` — works everywhere but changes role semantics for every host.

## Interim mitigation

billion-context PR ranxianglei/billion-context#356 hoists mid-conversation system/developer messages to the leading prefix at the OpenAI wire layer (both `prepareOpenai` and the compress-loop adapter), so proxy users are covered without a kernel release. This issue tracks the kernel-level root fix so all hosts benefit.

ranxianglei:
Fixes #170.

## Problem

`prune()` inserted a rendered summary (`role: "system"`) at the block's earliest covered-message index. Whenever a compress range started after the head — the normal "fold the old stuff, keep the tail" pattern — the summary landed **mid-conversation** (a system message after a user/assistant message). `coreToOpenai` serializes `role: "system"` verbatim at its position, so strict OpenAI-compatible backends reject every subsequent request with `400 System message must be at the beginning`, stranding the session (reproduced on sglang; ranxianglei/billion-context#355).

## Fix

**`src/prune.ts`** — clamp every summary anchor to the first non-system message so summaries always land in the leading system prefix. Clamping to the first *non-system* (not first *user*) message also covers hosts that lead with a non-user, non-system message. Summaries keep their chronological order via an unclamped `orderKey`, so multi-summary placement is unchanged apart from moving into the prefix.

**`src/boundaries.ts` / `src/compress.ts`** — moving the summary out of the range broke block detection for a block whose raws were all pruned (its summary was the only in-range representation), which regressed multi-tier distillation over a pruned region. `blockVisibleInRange` now falls back to the block's content refs when no raw survives in the view. The fallback is gated on `rawIndex === null` so a block whose anchor raw is merely outside the range (but present) is not wrongly treated as nested — that gate is what keeps the "covered by non-nested block" validation rejecting correctly.

## Tests

- 2 new regression tests pin the wire-validity invariant (no system message after a non-system message), including the leading-system-message case.
- 3 existing tests updated: their exact summary-position assertions reflected the old (buggy) mid-conversation placement.
- Full suite: 542 pass. typecheck + build clean.

Note: the issue's minimal proposal (`insertAt: Math.min(earliest ?? 0, firstUserIndex)`) is correct for placement but, on its own, regresses block consumption (the `promote-after-prune` "m-ref range spanning a pruned region" case) — the content-ref fallback here is required to keep distillation working.

### DIFF
--- a/src/boundaries.ts
+++ b/src/boundaries.ts
@@ -75,6 +75,11 @@ export interface ResolvedRange {
   boundaryKind: BoundaryKind;
   protectedGaps: number[];
   snappedBoundaries: string[];
+  // Numeric ref bounds of message-ref boundaries (null when either boundary is
+  // a block ref). Lets block detection fall back to a block's content refs when
+  // its raws are pruned and its summary sits in the leading system prefix.
+  refStart: number | null;
+  refEnd: number | null;
 }
 
 export function resolveBoundaries(
@@ -115,6 +120,14 @@ export function resolveBoundaries(
     [startIndex, endIndex] = [endIndex, startIndex];
   }
 
+  const bothMessageRefs = start.kind === "message" && end.kind === "message";
+  const refStart = bothMessageRefs
+    ? Math.min(start.numericId, end.numericId)
+    : null;
+  const refEnd = bothMessageRefs
+    ? Math.max(start.numericId, end.numericId)
+    : null;
+
   const messageIds: string[] = [];
   for (let index = startIndex; index <= endIndex; index++) {
     const message = input.messages[index];
@@ -131,7 +144,17 @@ export function resolveBoundaries(
   const nestedBlockIds: string[] = [];
   const nestedSeen = new Set<string>();
   for (const block of activeBlocks(input.state)) {
-    if (blockVisibleInRange(block, indexByMessageId, startIndex, endIndex)) {
+    if (
+      blockVisibleInRange(
+        block,
+        indexByMessageId,
+        startIndex,
+        endIndex,
+        input.state.messageRefs.byRaw,
+        refStart,
+        refEnd,
+      )
+    ) {
       if (!nestedSeen.has(block.blockId)) {
         nestedSeen.add(block.blockId);
         nestedBlockIds.push(block.blockId);
@@ -149,6 +172,8 @@ export function resolveBoundaries(
     boundaryKind,
     protectedGaps,
     snappedBoundaries,
+    refStart,
+    refEnd,
   };
 }
 
@@ -314,12 +339,19 @@ export function visibleBlockAnchor(
 // (rendered summary or earliest surviving raw) falls inside it. The summary
 // alone is not sufficient: when a block covers the session's first user
 // message, prune keeps that one raw and inserts the summary BEFORE it, so the
-// summary index can sit outside a range that still contains the raw.
+// summary index can sit outside a range that still contains the raw. The
+// content-ref fallback covers the inverse: a block whose raws are ALL pruned
+// (none survive in the view) while its summary sits in the leading system
+// prefix (also outside the range) — detected by its raw refs landing inside
+// the requested message-ref span.
 export function blockVisibleInRange(
   block: CompressionBlock,
   indexByMessageId: Map<string, number>,
   startIndex: number,
   endIndex: number,
+  byRaw: Record<string, string> = {},
+  refStart: number | null = null,
+  refEnd: number | null = null,
 ): boolean {
   const summaryIndex = indexByMessageId.get(summaryMessageId(block.blockId));
   if (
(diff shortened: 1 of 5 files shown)


ranxianglei:

[bot] **Independent review — approved.** ✅ (second-agent review per AGENTS.md; verified everything myself on branch `2026-08-30_prune-summary-anchor-wire` @ 4cf4c00, which sits directly on current master `bed2501` with no divergence and touches only the 5 intended files, no version bump.)

**Checks I ran:**
- `npm run typecheck` — clean
- `npm test` — **542 pass / 0 fail** (matches the PR claim)
- `npm run build` — clean
- **Fallback necessity**: reverted only `src/boundaries.ts` + `src/compress.ts` to master (keeping the `prune.ts` fix) and re-ran the suite — `promote-after-prune: m-ref range spanning a pruned region consumes the block without leaking synthetic ids` fails exactly as described, confirming the content-ref fallback in `blockVisibleInRange` is required, not optional. Restored; suite green again.
- **End-to-end wire check** (`prune` → `coreToOpenai` on a mid-conversation compress range): core roles come out `system(acp_summary_b1) user user assistant`, wire roles `system user user assistant` — the exact inversion of the old `user system user assistant` that triggered the sglang 400.

**Code-review notes (all non-blocking):**
1. **Clamping is robust.** `firstNonSystemIndex` ≤ `firstUserIndex` whenever a user message exists (the first user message is itself non-system), so summaries always land ahead of the preserved first-user raw; the degenerate all-system view falls back to index 0; multi-summary ordering via the unclamped `orderKey` preserves chronology when several anchors fold onto the same prefix slot; re-pruning an already-pruned view stays idempotent (existing-summary path is clamped too, which also self-heals views cached from pre-fix versions).
2. **The `rawIndex === null` gate is the right conservatism.** New detection is a strict superset of the old: surviving-raw blocks are decided exactly as before (raw/summary index check), and the content-ref fallback only fires when a block has *no* visible raw — precisely the case the old code lost. Without the gate, a block whose surviving raw sits before the range but whose other refs fall inside it would be silently consumed, folding coverage (`m5..m9`-style) that the model's m-ref summary doesn't describe — the gate correctly prevents that and keeps the livelock-guard rejection working.
3. **`refStart`/`refEnd` from the parsed (pre-snap) refs is the correct semantics** — the fallback asks "does this block's content overlap the span the model requested", which is ref-space, not view-index-space. Reversed boundaries are handled (`Math.min`/`Math.max`).
4. Minor observation: in the adjusted-range re-scan (`src/compress.ts` ~line 700), the fallback uses the *original* ref span, not the pair-extended one — a block whose content overlaps only the extension portion (and whose raws are all pruned) won't be detected there. This matches pre-PR behavior (summary-at-0 was equally out of range), so it's not a regression, just slightly conservative.
5. API changes are backward-compatible: `ResolvedRange.refStart/refEnd` are additive; `blockVisibleInRange`'s new params are optional with defaults; no other consumer depends on summary position (`sync.ts` checks presence only).
6. Test updates are legitimate: the 3 changed assertions encoded the old (buggy) mid-conversation placement, and the 2 new tests pin the wire-validity invariant including the leading-system-message case.

Ready for a human to merge — per AGENTS.md I can't merge PRs myself.

ranxianglei:
Closing as superseded: master `src/prune.ts:92-96` already implements exactly this — rendered summaries anchor at their existing position (`summaryIndexById`) "instead of jumping to index 0 when the raw ids are no longer in the input".

PR #173, opened 2026-08-31
fix(persist): fresh temp name per flushSync retry attempt

ranxianglei:
Hardening follow-up to #172 for the Windows EPERM storm (#362 in billion-context).

**Problem.** `flushSync` resolved ONE temp file name before the retry loop and reused it for every attempt. On Windows, a failed `unlinkSync` cleanup can leave that name *delete-pending*; creating over a tombstoned name fails immediately — so after the first failed attempt, every remaining attempt died at `writeFileSync` before the rename was even retried. The retry loop was effectively a single attempt in exactly the scenario it exists for.

**Fix.** `tempPath()` is now called inside the loop — one fresh name per attempt (the `pid` + monotonic `seq` suffixes keep them collision-free).

- Tests: 549 pass / 0 fail (new: distinct temp names per attempt + spill still lands)
- typecheck clean, build ok
- No API change; default retry policy unchanged


ranxianglei:

[bot] Reviewed the fix and verified it locally. The change itself is correct — I have one blocker I've already cleaned up.

## The fix (correct ✅)

`src/persist/store.ts:218-223` — moving `const tmp = this.tempPath(file)` from before the retry loop to inside it is exactly right. `tempPath()` (`src/persist/store.ts:434-436`) builds the name from `process.pid` + a monotonic `this.tmpSeq++`, so every attempt now gets a guaranteed-distinct name. That kills the exact failure mode you describe: a Windows delete-pending tombstone from a failed `unlinkSync` can no longer poison every subsequent attempt, because each attempt writes to a fresh name. The `catch` block's `unlinkSync(tmp)` also now cleans up the *current* attempt's temp, which is the correct pairing.

Verified locally:
- `npm run typecheck` — clean
- `npm test` — **549 pass / 0 fail**
- `npm run build` — success

The new test (`tests/persist.test.ts`) is well-targeted: it mocks `renameSync` to always throw EPERM, records every `writeFileSync` path, and asserts 3 temp writes that are all distinct (`new Set(tmps).size === 3`) plus the `.fb.json` spill landing. Before the fix that `Set` size would be 1, so it directly pins the regression. Good.

## Blocker I found & fixed: 3 junk files committed

The original commit (`ef1954a`) accidentally bundled three npm-cache artifacts under a literal `false/` directory:
- `false/_cacache/content-v2/sha512/fa/84/c372…c54` (109,894 B)
- `false/_cacache/index-v5/54/8c/fffb…ff2` (698 B)
- `false/_update-notifier-last-checked` (0 B)

These are not part of the fix. The `false/` prefix suggests npm ran with a cache path that resolved to the string `false` (e.g. `npm_config_cache=false` / `--cache false`), which created a real directory named `false` in the repo root.

I removed them in a follow-up commit `6b2418a` (`chore: drop accidentally committed npm cache artifacts (false/ dir)`) and pushed to the branch. The PR diff vs `master` is now just the two intended files:
- `M src/persist/store.ts`
- `M tests/persist.test.ts`

## Status

Ready for merge from my side. I can't merge PRs myself (system rules) — please merge when you're happy: https://github.com/ranxianglei/acp-kernel/pull/173

PR #181, opened 2026-08-31
fix(sync): keep host-expanded blocks deactivated across syncBlocks

ranxianglei:
## Problem

A block the host explicitly **decompressed** (`expanded`) was unconditionally re-activated by `syncBlocks` on the next turn: the client re-sends its full history, so the block's raw ids / summary id are present, and `syncBlocks` set `block.active = true` for every non-consumed block. This silently **re-folded the already-restored messages** — the model paid for the summary *and* the restored full text in the same round, and (with the host's content cache already deleted) a second `decompress` permanently degraded to the summary.

Root cause: the "user expanded this" intent had **no persistent trace** in the kernel — `deactivateBlock()` was wiped on the next `processTurn`.

## Fix

- `types.ts`: add `CompressionBlock.expanded?: boolean` — host-set, meaning "the user explicitly decompressed this block, so its deactivated state is intentional".
- `sync.ts`: in the activity pass, skip re-activation for `expanded` blocks (keep `active = false`). Blocks deactivated by other means (orphan GC, tier distillation, consumed-by-parent) keep the legacy resurrection behavior — the flag is the only thing that changes.
- `tests/sync-config.test.ts`: two cases — an expanded+deactivated block stays deactivated when its messages are present (and is not reported as orphaned); a non-expanded deactivated block is re-activated (legacy behavior preserved).

Back-compat: `expanded` is optional and defaults to absent/false, so existing persisted states are unaffected.

## Pre-flight

- `npm run typecheck` — clean
- `npm test` — 568 pass, 0 fail
- `npm run build` — success

This is a content PR. A follow-up release PR (v0.0.49) will be cut from master once this lands, per the release workflow.

### DIFF
--- a/src/sync.ts
+++ b/src/sync.ts
@@ -65,6 +65,14 @@ export function syncBlocks(
       block.active = false;
       continue;
     }
+    // Host-set `expanded` = the user explicitly decompressed this block, so its
+    // deactivated state is intentional. Re-activating it here would re-fold the
+    // already-restored messages next turn (double cost + lost originals). Keep
+    // it inactive; a fresh compress of the same range creates a NEW block.
+    if (block.expanded) {
+      block.active = false;
+      continue;
+    }
     block.active = true;
     // A block whose raw messages were replaced by its rendered summary
     // (pruned view) is still present — the summary IS the block's visible
--- a/src/types.ts
+++ b/src/types.ts
@@ -31,6 +31,12 @@ export interface CompressionBlock {
   survivedCount: number;
   generation: BlockGeneration;
   active: boolean;
+  /** Host-set: the user explicitly decompressed (expanded) this block, so its
+   *  deactivated state is intentional and must survive syncBlocks re-activation.
+   *  Absent/false for blocks deactivated by other means (orphan GC, tier
+   *  distillation, consumed-by-parent) — those keep the legacy resurrection
+   *  behavior. */
+  expanded?: boolean;
   durationMs?: number;
   compressCallId?: string;
   startRef?: string;
--- a/tests/sync-config.test.ts
+++ b/tests/sync-config.test.ts
@@ -50,6 +50,25 @@ test("syncBlocks leaves blocks intact when at least one message remains", () => 
   assert.equal(result.state.blocks[0]!.active, true);
 });
 
+test("syncBlocks keeps expanded (user-decompressed) blocks deactivated when present", () => {
+  const state = createInitialState();
+  state.blocks.push(
+    makeBlock({ blockId: "b1", effectiveMessageIds: ["a", "b"], active: false, expanded: true }),
+  );
+  const result = syncBlocks([msg("a"), msg("b")], state);
+  assert.equal(result.state.blocks[0]!.active, false, "expanded block must stay deactivated");
+  assert.deepEqual(result.deactivated, [], "expanded block is not reported as orphaned");
+});
+
+test("syncBlocks re-activates a deactivated (non-expanded) block when its messages return", () => {
+  const state = createInitialState();
+  state.blocks.push(
+    makeBlock({ blockId: "b1", effectiveMessageIds: ["a", "b"], active: false }),
+  );
+  const result = syncBlocks([msg("a"), msg("b")], state);
+  assert.equal(result.state.blocks[0]!.active, true, "legacy resurrection preserved for non-expanded blocks");
+});
+
 test("syncBlocks does not mutate input state", () => {
   const state = createInitialState();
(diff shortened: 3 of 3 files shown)


ranxianglei:

[bot] 🏷 Reviewed the PR on branch `2026-08-31_expanded-block-sync` (commit `0240d90`). **LGTM** — correct, minimal, matches the stated intent. I re-ran the pre-flight on the branch rather than trusting the description.

**Pre-flight (re-run, green):**
- `npm run typecheck` — clean
- `npm test` — 568 pass, 0 fail
- `npm run build` — success

**Change review:**
- `src/types.ts` — `expanded?: boolean` on `CompressionBlock` (optional, host-set). Back-compat confirmed: absent ⇒ falsy, so old persisted states are unaffected. The `syncBlocks` deep-clone copies it via the `...block` spread, so it survives into the returned state.
- `src/sync.ts:68-75` — placement is correct: after the consumed-by-parent check (consumed wins) and before the legacy `block.active = true`. An expanded block is kept `active = false` and is **not** pushed to `deactivated` (right — it's intentionally deactivated, not orphaned).
- `tests/sync-config.test.ts` — both cases are meaningful: expanded+deactivated stays deactivated and is not reported orphaned; non-expanded deactivated is re-activated (legacy preserved).

**Interaction analysis (no regressions):**
- `advanceSurvival` (`src/state.ts:62`) skips inactive blocks → no interaction.
- `coveredMessageIds` (`src/state.ts:45`) now excludes the expanded block's messages → they're uncovered/visible. That *is* the fix: the restored raw text is the only representation, so no summary+raw double cost.
- `hideConsumedCompressCalls` (`src/hide-consumed.ts:46`) now hides the expanded block's compress call (block is inactive). Correct consequence — the raw content is already visible, the stale call is just noise. (Behavior change vs. pre-PR, where `syncBlocks` re-activated the block and kept the call visible.)

**Notes (non-blocking):**
1. **Host wiring is the prerequisite.** The kernel change is a no-op until the host (pai-acp) sets `expanded: true` when the user decompresses a block *and* persists it across turns. That's a separate change in the adapter repo; this PR is the kernel-side prerequisite. Worth confirming the host PR is tracked so the fix is end-to-end.
2. **Edge case (acceptable, untested):** an expanded block whose raw messages are all gone (e.g. host-truncated) stays deactivated and is *not* reported as orphaned (the `continue` skips the `deactivated.push`). It becomes a permanent tombstone — consistent with the GC-free design and the "intentionally deactivated" semantics. Flagging for completeness, not a blocker.

Not merging — that's human-only. Ready to land once the host-side wiring is confirmed.

ranxianglei:
[bot] 🏷 Re: "Host wiring is the prerequisite" — confirming the host PR is tracked.

The host for this fix is **billion-context** (the `bili` proxy), not pai-acp. The host-side wiring is already implemented and verified locally (commit `12c9b38`, branch `2026-08-31_decompress-expanded`):
- `resolveDecompress` sets `block.expanded = true` when it deactivates a block on successful decompress (persisted in the session state).
- The loop-path decompress fallback now scans `compressMessages ?? messages`, so it recovers the same originals the stream path does.

That billion-context PR will be opened as soon as this change ships as **0.0.49** (billion-context pins acp-kernel to an exact version, so it cannot land before the 0.0.49 publish).

The kernel change here is safe to merge standalone: `expanded` is optional and host-set, so any host/state that does not set it keeps the exact legacy behavior (resurrection for non-consumed blocks). No end-to-end regression risk from merging this first.

Sequence: merge this → cut 0.0.49 release → CI publish → open the billion-context host-wiring PR.

ranxianglei:
**[设计反转 — 建议关掉，重做为无状态版]**

维护者已拍板（billion-context #400 新评论）：**decompress 必须是无状态纯检索**——不 deactivate、不改视图、不删缓存。"用户解压过"从此不是需要记录的状态。

因此本 PR 的 `expanded` 字段及其 syncBlocks 门控不再需要：无状态语义下块永远 `active`，复活问题不存在。

**建议**：close 本 PR（实现本身没问题，只是语义被推翻），改为 `decompress()` 去掉 `deactivateBlock` 调用（`src/decompress.ts:51`）的无状态版本。bili 侧配套 PR 同步重做。

ranxianglei:
**更正上文**：内核无需任何改动 —— `core.decompress` 本来就是纯查块，`collectBlockContent` 本来就是无状态检索原语，deactivate 从来不是内核内部行为（是宿主调用）。**本 PR 直接 close 即可**，不需要无状态重做版。bili 侧修复见 billion-context #400 新评论。

Issue #187, opened 2026-09-03
wire codecs: silent image loss — openai single remote-URL image dropped; responses image-only user item dropped


ranxianglei:
来源: ranxianglei/billion-context#488 分析 multimodal payload accounting 时发现（billion-context PR #489 review）

Two silent image-loss gaps in the wire codecs (v0.0.47, dist/wire/index.js). Both drop image content from rebuilt request bodies without any error.

## 1. openaiToCore: single non-data-URL image gets no sidecar → dropped on re-encode

`openaiToCore` case "user" (dist/wire/index.js:243-257):

```js
...imgs.length === 1 && firstParsed ? { rawOpenaiContent: imgs[0], imageMediaType: firstParsed.mediaType, imageBase64: firstParsed.base64 }
 : imgs.length > 1 ? { rawOpenaiContentParts: imgs } : {}
```

For a user message containing exactly ONE image whose url is a remote http(s) URL (not a data: URL): `firstParsed = parseDataUrl(url)` is undefined → first branch false; `imgs.length > 1` false → **no sidecar field at all**. `coreToOpenai` then rebuilds content from text alone — the image is silently lost from the forwarded request. Data URLs and multi-image messages are preserved; only the single-remote-URL shape drops.

Repro: `{messages:[{role:"user",content:[{type:"text",text:"t"},{type:"image_url",image_url:{url:"https://example.com/a.png"}}]}]}` through `openaiToCore` → `coreToOpenai`: output content has no image part.

Suggested fix: keep the raw part whenever there is exactly one image, splitting out base64 only when it is a data URL:

```js
...imgs.length === 1
  ? { rawOpenaiContent: imgs[0], ...(firstParsed ? { imageMediaType: firstParsed.mediaType, imageBase64: firstParsed.base64 } : {}) }
  : imgs.length > 1 ? { rawOpenaiContentParts: imgs } : {}
```

`coreToOpenai` already re-emits `rawOpenaiContent` verbatim when present (dist/wire/index.js:355-367), so no encode-side change is needed.

## 2. responsesToCore: image-only user item produces no core message → dropped from rebuilt input

[...]

PR #190, opened 2026-09-03
fix(wire): skip malformed image_url:null parts in openaiToCore (#188)

### LINKED ISSUE #188: openaiToCore 崩溃：image_url 为 null 的 image_url part 触发 TypeError
来源: ranxianglei/acp-kernel#187 修复过程中 review 时发现（独立代码审查，已验证 master v0.0.49 上字节级可复现）

## 问题

`openaiToCore` 遇到 `image_url` 字段为 `null` 的 image part 时直接抛异常（src/wire/openai.ts:271）：

```js
const url = imagePart.image_url.url;
```

复现（master v0.0.49）：

```js
openaiToCore({ messages: [{ role: "user", content: [{ type: "image_url", image_url: null }] }] });
// TypeError: Cannot read properties of null (reading 'url')
//   at allImageParts (src/wire/openai.ts:271)
```

## 影响

- **预存在问题**，与 #187 的修改无关（该行在 HEAD 上逐字节相同），#187 的 PR 不触碰它。
- billion-context proxy 模式：任何发送 `{type:"image_url", image_url:null}`（畸形但 JSON 合法）的客户端请求都会在 `*ToCore` 内抛出未捕获异常，整个 turn 失败或代理崩溃，而不是优雅降级。

## 建议修复

在 `allImageParts` 中加空值守卫：`image_part.image_url?.url` 非 string 时跳过该 part（按非图片 part 处理），其余收集语义保持不变。

ranxianglei:
## Problem

`openaiToCore` throws an uncaught `TypeError` on an image part whose `image_url` is `null`:

```js
openaiToCore({ messages: [{ role: "user", content: [{ type: "image_url", image_url: null }] }] });
// TypeError: Cannot read properties of null (reading 'url')
```

Malformed but JSON-valid input. In billion-context proxy mode, any client request carrying such a part crashes the whole turn (or the proxy) instead of degrading gracefully.

**Pre-existing**, unrelated to #187 (that line is byte-identical at HEAD; #187's diff does not touch it). Found during independent review while working on #187.

## Root cause

`allImageParts` (src/wire/openai.ts) gates on `"image_url" in p`, which is `true` even when the value is `null`, then reads `imagePart.image_url.url` → `null.url` → TypeError. The gate checks key *presence*, not value validity.

## Fix

Widen the narrowing cast to allow null/absent `image_url` and optional-chain the read so such parts degrade to non-image parts (skipped):

```diff
-        const imagePart = p as { image_url: { url?: unknown } };
-        const url = imagePart.image_url.url;
+        const imagePart = p as { image_url?: { url?: unknown } | null };
+        const url = imagePart.image_url?.url;
         if (typeof url === "string" && parseDataUrl(url)) out.push(p as OpenAIImagePart);
```

The existing `typeof url === "string"` guard already skips non-string urls, so no other collection semantics change. This transitively protects `openai.ts:70` (`firstImg.image_url.url`) since `firstImg` now only ever comes from well-formed parts.

Sibling audit: `responses.ts` already guards `typeof part.image_url === "string"` (safe); no other vulnerable sites found.

## Verification

- Repro on v0.0.49 pre-fix: throws `TypeError: Cannot read properties of null (reading 'url')`.
- Post-fix: no throw; malformed part skipped, user message preserved (empty text, no image sidecar).
- Regression test added (`tests/wire-bili-message-roundtrip.test.ts`, case 5c) — confirmed to **fail on unfixed code**, pass post-fix.
- `npm run typecheck` clean; `npm test` → **573 pass / 0 fail**.

Fixes #188

### DIFF
--- a/src/wire/openai.ts
+++ b/src/wire/openai.ts
@@ -259,8 +259,8 @@ function allImageParts(content: OpenAIMessage["content"]): OpenAIImagePart[] {
         if (!("type" in p) || p.type !== "image_url" || !("image_url" in p)) continue;
         // The union's index-signature member leaves image_url as `unknown`, so
         // narrow via a named const before reading the url.
-        const imagePart = p as { image_url: { url?: unknown } };
-        const url = imagePart.image_url.url;
+        const imagePart = p as { image_url?: { url?: unknown } | null };
+        const url = imagePart.image_url?.url;
         if (typeof url === "string" && parseDataUrl(url)) out.push(p as OpenAIImagePart);
     }
     return out;
--- a/tests/wire-bili-message-roundtrip.test.ts
+++ b/tests/wire-bili-message-roundtrip.test.ts
@@ -191,6 +191,21 @@ test("openai: user message with multiple image_url parts round-trips ALL images"
     assert.ok(typeof text?.text === "string" && text.text.startsWith("compare these?"), "text preserved");
 });
 
+// 5c. Malformed but JSON-valid image part (image_url:null) must degrade to a
+//     non-image part instead of throwing — previously raised TypeError reading
+//     null.url inside allImageParts, crashing the whole turn in proxy mode.
+test("openai: malformed image_url:null part is skipped without throwing", () => {
+    const body = {
+        messages: [
+            { role: "user", content: [{ type: "image_url", image_url: null }] },
+        ],
+    } as unknown as OpenAIRequestBody;
+    const { msgs } = openaiToCore(body);
+    assert.equal(msgs.length, 1);
+    assert.equal(msgs[0]?.imageMediaType, undefined, "null image_url degrades to non-image");
+    assert.equal(msgs[0]?.imageBase64, undefined);
+});
+
 // 6. Responses API input_image round-trips (image was previously dropped).
 test("responses: user input_image is restored via rawResponsesItem", () => {
     const body: ResponsesRequestBody = {


ranxianglei:
> 🤖 ework agent · qwen3.8-27b

## Review: PR #190 — fix(wire): skip malformed image_url:null parts in openaiToCore (issue #188)

**Verdict: ✅ approve** — bug verified real, root cause correctly identified, fix is minimal and correct, all pre-flight checks green.

### 1. Reproduction (verified independently)

Repro'd on master v0.0.49 (`84adf12`, the PR base) with the exact snippet from the issue:

```js
openaiToCore({ messages: [{ role: "user", content: [{ type: "image_url", image_url: null }] }] });
// → TypeError: Cannot read properties of null (reading 'url')  at allImageParts (src/wire/openai.ts:263)
```

Confirmed pre-existing and unrelated to #187 — the vulnerable line is byte-identical at HEAD.

### 2. Root-cause layering

The report's analysis is accurate: `allImageParts` gates on `"image_url" in p` (key *presence*), which is true for `{ type: "image_url", image_url: null }`, then dereferences `imagePart.image_url.url` → `null.url`. The gate checks presence, not value validity — that's the root cause, not a symptom, and the fix lands exactly there. No deeper layering issue found.

### 3. Fix evaluation

- Widening the narrowing cast to `{ image_url?: { url?: unknown } | null }` + optional-chaining `imagePart.image_url?.url` is the right shape: the existing `typeof url === "string"` guard already skips null/non-string urls, so collection semantics are unchanged for all well-formed inputs.
- The fix preserves the invariant that `allImageParts` only returns well-formed parts (`url` is a string AND `parseDataUrl(url)` truthy), which transitively protects `src/wire/openai.ts:70` (`firstImg.image_url.url`) as claimed. Verified by reading the data flow; no other reader of OpenAI chat `image_url` exists besides line 70 (write-side line 201 builds its own object).
- Alternative considered: tightening the gate to `p.image_url != null`. Equivalent outcome, slightly less idiomatic than optional chaining here. The chosen approach is fine.

### 4. Independent sibling audit

- `src/wire/responses.ts:209` — guards `typeof part.image_url === "string"` before use (and `image_url` is typed `string` there, not an object). Safe, matches the PR's claim.
- `src/wire/anthropic.ts` — no image-content handling at all. Safe.
- `src/wire/formats.ts:22` — key-presence checks only, never reads values. Safe.
- No other vulnerable sites found.

### 5. Verification (re-ran everything on `e20fda2`)

| Check | Result |
|---|---|
| New regression test (case 5c) on **unfixed** code | ❌ fails with `TypeError: Cannot read properties of null (reading 'url')` (as claimed) |
| Same test on fixed branch | ✅ pass |
| `npm run typecheck` | ✅ clean |
| `npm test` | ✅ **573 pass / 0 fail** (matches claim) |
| `npm run build` | ✅ success |

Extra sanity beyond the PR's test: mixed input `[malformed part, valid data-url image]` still collects the valid image (`imageBase64` set, mediaType `image/png`) and doesn't throw; round-trip `coreToOpenai` on the degraded message yields plain string content — graceful degradation confirmed end-to-end.

### Minor observations (non-blocking)

- The new test could also assert `rawOpenaiContent` / `rawOpenaiContentParts` are undefined and cover the mixed malformed+valid case — current assertions are sufficient for regression purposes, just noting.
- Degraded messages keep `"\n"`-joined text from `stringContent` (e.g. text part + skipped image → `"hello\n"`). Pre-existing behavior for any non-text part type, unchanged by this fix — no action needed.

Per AGENTS.md, merge stays with you (agents must not merge). With this review + one more agent sign-off, it's ready.

PR #199, opened 2026-09-04
fix: gate pressure-band nudges on a minimum benefit floor (#198)

### LINKED ISSUE #198: [Enhancement] decideNudge pressure band injects micro/zero-yield EMERGENCY nudges — no minimum-benefit gate
来源: ranxianglei/billion-context-pi#289 分析 [Bug] Token 计数与压缩后状态不同步，导致低收益及零收益的连续压缩 时发现

## 现象 (observed in production, v0.0.48)

In a long multi-block session the following EMERGENCY nudges fired repeatedly:

- `EMERGENCY T2 distill: max pending 232 (T1 effective 0, T2 232, T3 0), usage 128%`
- `EMERGENCY T3 condense: max pending 120 ..., usage 128%`

Each resulting T2/T3 "compression" rewrote a summary at nearly the same size (~0 tokens reclaimed), which reset the nudge baselines on success, so the next LLM call re-injected the same EMERGENCY nudge — a continuous low/zero-yield compression loop.

## Root cause in kernel

`decideNudge` (src/compress.ts, v0.0.48): when `usage >= nudge.maxContextLimitPct || usage >= nudge.emergencyThresholdPct`, the pressure branch picks the tier with the MAX pending amount and injects whenever `bestPending > 0`. There is no lower bound on how small pending may be — pending of 232 or 120 tokens still triggers an EMERGENCY injection that asks the model to "distill NOW".

The loop is amplified when the host-supplied `tokenCount` is inflated relative to the real sent view (that adapter-side bug is fixed in ranxianglei/billion-context-pi#293): usage stays pinned above the emergency threshold even after compressions, so the pressure band never releases and every turn re-injects.

## Suggested fix

In the pressure band, require a minimum benefit before injecting, e.g.:

- `bestPending >= minBenefitTokens`, where `minBenefitTokens` scales with the window (e.g. `max(5000, round(limit * 0.01))`), or at least `>= compress.minCompressRange` expressed in tokens; otherwise suppress the nudge (and say why — usage high but nothing worth compressing) instead of offering a sub-threshold rewrite.

This makes the pressure band robust regardless of how accurate the host's tokenCount is: a tiny pending can never again masquerade as an actionable emergency.

## Why not fix adapter-side only

The adapter can no longer inflate counts (#293), but any host that over-reports tokenCount would reproduce the same loop against the current kernel gate. The gate belongs next to the decision.

ranxianglei:
Fixes #198.

## Problem
The pressure band in `decideNudge` (usage >= maxContextLimitPct / emergencyThresholdPct) picked the argmax tier and injected whenever `bestPending > 0`. In production (v0.0.48) this fired `EMERGENCY T2 distill: max pending 232 ... usage 128%` and `EMERGENCY T3 condense: max pending 120 ...` repeatedly: each rewrite reclaimed ~0 tokens, each success reset the nudge baselines, and the still-hot band re-injected the same EMERGENCY nudge every turn - a continuous zero-yield compression loop.

## Fix
- Pressure band now injects only when `bestPending >= minPressureBenefit`, default **`max(5000, round(modelContextLimit * 0.01))`** (5K floor up to 500K windows, then 1% of window: 2K @ 200K, 10K @ 1M). A tiny pending can no longer masquerade as an actionable emergency regardless of how accurate the host tokenCount is.
- New optional config knob `nudge.minPressureBenefitTokens` for hosts that need to tune it; explicit `0` restores the legacy any-pending behavior. Validated in `validateConfig` (finite, >= 0).
- When suppressed, the reason states why: usage high but max pending below the benefit floor - instead of offering a sub-threshold rewrite. The zero-pending reason is unchanged; `truncate.threshold` remains the independent last resort.
- `NudgeBreakdown` now exposes `minPressureBenefit` for adapters/debugging.

## Scope
Pressure band only. The count-triggered T2/T3 paths (`tier2Trigger`/`tier3Trigger`) deliberately fire on block count regardless of summary token mass (LSM-style compaction) and are untouched, as is the #194 first-sight mass bypass.

## Tests
`tests/nudge-pressure-benefit.test.ts` (new) mirrors the production pendings at exact values:
- EMERGENCY + only 232-token T2 pending -> suppressed, reason explains the shortfall, second turn stays suppressed with nothing stamped (loop cannot re-arm)
- EMERGENCY + only 120-token T3 pending -> suppressed
- pending exactly at the floor (5000) -> still injects (boundary inclusive)
- 1M window -> floor scales to 10000 (6000 pending suppressed; 10000 injects)
- override 100 honored (232 injects); override 0 restores legacy behavior

All three suppression tests failed against unmodified master (captured before implementing); after the fix: **571/571 tests pass**, `tsc --noEmit` clean, build clean.

Three pre-existing regression tests (`tests/regression-bugfixes.test.ts`) used pressure injection with micro pending purely as scaffolding for unrelated assertions (ref assignment, input-state immutability, pipeline composition). They opt out via the documented `minPressureBenefitTokens: 0` escape hatch, keeping them testing what they always tested.

### DIFF
--- a/src/compress.ts
+++ b/src/compress.ts
@@ -1069,6 +1069,22 @@ function resolveAdaptiveGrowth(
   );
 }
 
+/** Minimum reclaimable tokens for a pressure-band nudge to be worth injecting.
+ *  A sub-threshold rewrite (e.g. re-distilling a 232-token summary at near-
+ *  equal size) resets the nudge baselines on success, re-arming the still-hot
+ *  band next turn — a zero-yield loop (#198). Scales with the window so an
+ *  inflated host tokenCount can never make a tiny pending look actionable:
+ *  max(5000, round(limit × 0.01)); explicit 0 restores legacy any-pending. */
+function resolveMinPressureBenefit(
+  modelContextLimit: number,
+  nudge: NudgeConfig,
+): number {
+  return (
+    nudge.minPressureBenefitTokens ??
+    Math.max(5000, Math.round(modelContextLimit * 0.01))
+  );
+}
+
 /** Compressible amount for each tier. T1 = EFFECTIVE merged-range tokens —
  *  only ranges whose real char count >= minCompressRange count (avoids
  *  inflation from fragmentation; matches the apply-side gate, which counts
@@ -1115,6 +1131,7 @@ function decideNudge(input: NudgeInput): NudgeDecision {
   const usage = limit > 0 ? tokenCount / limit : 0;
 
   const nudgeGrowthTokens = resolveAdaptiveGrowth(limit, config.nudge);
+  const minPressureBenefit = resolveMinPressureBenefit(limit, config.nudge);
 
   const overLimit = usage >= config.nudge.maxContextLimitPct;
   const emergencyOverride = usage >= config.nudge.emergencyThresholdPct;
@@ -1165,6 +1182,7 @@ function decideNudge(input: NudgeInput): NudgeDecision {
   );
   let injectedTier: CompressionTier | null = null;
   let injectedReason = "";
+  let bestPending = 0;
   const t1Eff = tiers[1]?.pending ?? 0;
   const t2Pen = tiers[2]?.pending ?? 0;
   const t3Pen = tiers[3]?.pending ?? 0;
@@ -1203,7 +1221,6 @@ function decideNudge(input: NudgeInput): NudgeDecision {
       candidates.push(2, 3);
     }
     let best: CompressionTier | null = null;
-    let bestPending = 0;
     for (const t of candidates) {
       const p = tiers[t]?.pending ?? 0;
       if (p > bestPending) {
@@ -1211,7 +1228,10 @@ function decideNudge(input: NudgeInput): NudgeDecision {
         best = t;
       }
     }
-    if (best !== null && bestPending > 0) {
+    // Minimum-benefit gate (#198): a sub-threshold pending can never look
+    // actionable, no matter how hot the band is — otherwise every "success"
+    // resets the baselines and the same EMERGENCY nudge re-injects each turn.
+    if (best !== null && bestPending >= minPressureBenefit) {
       injectedTier = best;
       const label = emergencyOverride ? "EMERGENCY" : "OVER-LIMIT";
       injectedReason =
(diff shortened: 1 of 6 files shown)


ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] 🏷 **PR review — verified independently, approve (one non-blocking nit)**

Checked out the PR commit (`56027ee`, single commit over base `origin/master` = `003fbfa`, v0.0.51) and re-ran everything on real Node v22.23.2:

**Claims verified**
- On the PR commit: `tsc --noEmit` clean, **571/571 tests pass**, build clean — matches the description.
- Re-ran the new `tests/nudge-pressure-benefit.test.ts` against **unmodified base master**: 4 of 6 cases fail — the three suppression scenarios (232-token T2, 120-token T3, and the 6000 < 10000 half of the 1M-window test), plus the override-100 case which fails only on its `breakdown.minPressureBenefit === 100` assertion (field doesn't exist on master; its injection-behavior assert passes under legacy). So "three suppression tests failed pre-fix" holds for the suppression *scenarios*; strictly it's 4/6 cases. Either way the tests genuinely pin the bug, not scaffolding.
- The escape hatch is necessary, not cosmetic: running the base-master version of `tests/regression-bugfixes.test.ts` against the PR code fails exactly the three tests that opt out via `minPressureBenefitTokens: 0` (ref assignment, input-state immutability, pipeline composition) and nothing else.

**Scope & semantics**
- Diff touches only the pressure branch of `decideNudge`: gate `bestPending > 0` → `bestPending >= minPressureBenefit`, the suppression-reason split, and the breakdown field. Count-triggered T2/T3 paths and the #194 first-sight mass bypass are untouched. The zero-pending reason string is byte-identical to the old one.
- `resolveMinPressureBenefit`: `nudge.minPressureBenefitTokens ?? Math.max(5000, Math.round(limit * 0.01))`. Explicit `0` restores legacy any-pending behavior (argmax still requires pending > 0); a negative value degrades harmlessly to legacy even if validation is bypassed; `validateConfig` correctly flags non-finite / < 0.
- The suppression tradeoff is sound: when usage ≥ threshold but total compressible pending is under the floor there is genuinely little to reclaim, and `truncate.threshold` remains the independent last resort.

**Non-blocking nit (suggestion)**
`NudgeBreakdown` picks up `minPressureBenefit` through its index signature `[key: string]: number` (src/types.ts:279) rather than an explicit declaration. The interface's own doc comment says the fields are typed precisely so adapters get a compile-time signal when a key is renamed or removed — declaring `minPressureBenefit: number;` in the interface (src/types.ts:266) would keep the adapter-facing contract consistent with that intent. One line; feel free to fold it in or leave as-is.

**Release-note reminder** (already flagged on #198): behavior change under high usage — sessions with tiny pending will see a suppression reason instead of an EMERGENCY nudge until pending grows past the floor or truncation engages.

Verdict: approved from my side. Per AGENTS.md this needs the second-agent review and a human merge — I won't merge.

PR #201, opened 2026-09-05
fix: retry whole write cycle on transient ENOENT/ENOTDIR (CI Windows temp sweep)

ranxianglei:
## Problem

billion-context CI on `windows-latest` flakes chronically (seen on #540/#535/#539/#545/#548/#337-rerun, ~1 failure per PR, always healed by rerun). Two captured signatures, identical root:

```
[persist] write failed for debounce-1: ENOENT: no such file or directory, rename
  'C:\...\Temp\bili-persist-WavbaK\_unknown\.tmp-c6f17...-4432-0'
  -> 'C:\...\_unknown\c6f17....json'; SPILLOVER ALSO FAILED — data at risk

[persist] write failed for e2e-export-sess: ENOENT ... rename .../bili-e2e-export-2WEbQ0/openai/.tmp-...
  -> .../openai/127.0.0.1_f2f6....json; SPILLOVER ALSO FAILED — data at risk
```

`writeFile(tmp)` **succeeded**, then `rename` ENOENT (tmp vanished), then the single-shot spill write **also** ENOENT (directory gone). A third error in the same window shows `mkdir` failing `ENOTDIR`. Conclusion: something on GitHub Windows runners (periodic %TEMP% sweeper / Defender real-time scan) deletes freshly-created files/dirs milliseconds after creation. Real user machines with aggressive AV exhibit the same class of failure — this is not CI-only.

## Why the existing retry didn't help

`renameWithRetry` only retried **EPERM/EBUSY/EACCES** (lock semantics). **ENOENT rethrows immediately**, and the spill write had no retry at all — so a transient sweep drops the record entirely ("data at risk").

## Fix

- `writeInner` (async) and `flushSync`: retry the **whole write cycle** per attempt — `mkdir` (recreates a swept dir) + fresh `.tmp-` write + `rename` — for the transient set `EPERM|EBUSY|EACCES|ENOENT|ENOTDIR`. Existing exponential backoff/attempt policy unchanged (defaults: 6 attempts, ~1.5s window).
- The spill write gets its own transient retry window (both observed flakes had canonical AND spill failing together).
- `renameWithRetry` folded into the outer loop and removed.

Non-transient errors (e.g. ENOSPC, EROFS, EISDIR) still fail fast; permanent ENOTDIR (a blocker file on the path) now consumes the retry window (~1.5s with defaults) before spilling — acceptable, and failure-path tests in downstream use small retry configs anyway.

## Tests (+4, all mock-injected via `t.mock.method`)

- temp file vanishing before rename (rename ENOENT once) → healed, canonical written, no spill
- directory vanishing (writeFile ENOENT once) → recreated on retry
- persistent ENOENT → spills, data survives (mirrors the EPERM spill test)
- flushSync heals an ENOENT sweep the same way

## Pre-flight

- `npm run typecheck` ✓
- `npm test` 575/575 ✓
- `npm run build` ✓

Downstream: billion-context will bump its pinned `acp-kernel` after this ships, plus replace the fixed sleeps in its two flaky tests with deadline polls (belt and braces).

### DIFF
--- a/src/persist/store.ts
+++ b/src/persist/store.ts
@@ -75,13 +75,18 @@ export interface StateStoreOptions<T> {
      */
     validate?: (envelope: PersistedEnvelope<T>) => boolean;
     /**
-     * Rename-retry policy for Windows transient locks (EPERM/EBUSY/EACCES
-     * from AV scan, search indexer, SMB). Exponential backoff: the delay
-     * after attempt i is min(retryBaseMs * 2^i, retryMaxMs). Defaults
-     * (6 attempts, 50ms base, 1600ms cap) give a ~1.5s window — long enough
-     * for most AV locks to release, short enough not to stall a sync flush.
-     * When the window is exhausted the write spills to a side file (see
-     * spillPath) instead of dropping the data.
+     * Transient-write retry policy (Windows + hostile temp environments).
+     * The whole write cycle — mkdir, temp write, rename — is retried with
+     * exponential backoff when it fails with a TRANSIENT code: EPERM/EBUSY/
+     * EACCES (a lock held by AV scan, search indexer, SMB) or ENOENT/ENOTDIR
+     * (the temp file or its directory vanished between create and rename —
+     * observed on CI Windows runners where a sweeper deletes freshly
+     * created files under %TEMP%). The delay after attempt i is
+     * min(retryBaseMs * 2^i, retryMaxMs). Defaults (6 attempts, 50ms base,
+     * 1600ms cap) give a ~1.5s window per write path — long enough for most
+     * locks to release, short enough not to stall a sync flush. When the
+     * window is exhausted the write spills to a side file (see spillPath)
+     * instead of dropping the data.
      */
     retryAttempts?: number;
     retryBaseMs?: number;
@@ -94,10 +99,12 @@ export interface StateStoreOptions<T> {
  *
  * - atomic writes: temp file + rename, so a crash mid-write never leaves a
  *   truncated record (readers see either the old or the new file)
- * - rename retries with exponential backoff on Windows transient locks
- *   (EPERM/EBUSY/EACCES); when the window is exhausted the record spills to
- *   a side file (`<name>.fb.json`) instead of being dropped, so a lock held
- *   by AV/indexer/SMB never silently loses session data
+ * - whole-cycle retries with exponential backoff on transient fs failures:
+ *   Windows locks (EPERM/EBUSY/EACCES from AV/indexer/SMB) and vanishing
+ *   files/dirs (ENOENT/ENOTDIR — a sweeper deleting fresh temp files, seen
+ *   on CI Windows runners); when the window is exhausted the record spills
+ *   to a side file (`<name>.fb.json`) instead of being dropped, so a lock
+ *   held by AV/indexer/SMB never silently loses session data
  * - per-id serialization so concurrent writeNow calls never interleave
  *   temp-file names or reorders writes
  * - debounced scheduleSave coalesces bursts into one write; the record is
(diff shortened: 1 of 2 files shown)


ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] 🏷 **Review pass** — verified the bug against the pre-PR code, reviewed the diff line-by-line, and re-ran the full pre-flight locally.

**Duplicate screen:** not a duplicate — this extends the merged #172 (spill-on-rename-failure) and #173 (fresh temp name per flushSync attempt) rather than overlapping them.

**1. Bug verification.** Checked out `e371341` (single commit on current master `5fc164a`). Pre-PR code matches the description exactly:
- `renameWithRetry` retried only `EPERM|EBUSY|EACCES`; every other code — including ENOENT — rethrew on the first failure.
- The spill write was a single try/catch with zero retries, so a sweep that takes the directory drops the record entirely ("SPILLOVER ALSO FAILED — data at risk").
- The captured signatures (writeFile succeeds → rename ENOENT → spill ENOENT, plus a mkdir ENOTDIR in the same window) are consistent with transient deletion of freshly created files/dirs under `%TEMP%`, and the same class applies to aggressive AV on real machines — so the fix is correctly scoped to the kernel, not CI workarounds.

**2. Layer judgment.** Root cause is the transient vanishing of the temp file *or its directory* between create and rename; "data at risk" is the downstream symptom. The fix targets the root cause (whole-cycle retry incl. mkdir recreation + fresh tmp name, and giving the spill its own retry window) rather than papering over the symptom. Non-transient codes (ENOSPC/EROFS/EISDIR…) still break on the first attempt and fail fast — verified in `isTransientFsError`.

**3. Code review** (`src/persist/store.ts`):
- `writeInner` (async, ~L382–433) and `flushSync` (~L220–267) now both loop `mkdir → fresh .tmp- write → rename` per attempt over the transient set `{EPERM, EBUSY, EACCES, ENOENT, ENOTDIR}`; backoff policy untouched (defaults 6 attempts / 50ms base / 1600ms cap ≈ 1.5s window; constructor clamps `retryAttempts ≥ 1` so the loop always runs).
- Per-attempt fresh `tempPath()` preserves the #173 delete-pending-tombstone reasoning; orphan `.tmp-` cleanup moved inside the loop (per attempt) in both paths. Success path unchanged: `discovered.set → clearFailure → removeSpill`.
- `renameWithRetry` removed cleanly — zero leftover references (grepped src + tests).
- Sync/async structures are mirror-correct (Atomics.wait vs setTimeout sleep, same predicate, same early-break conditions).

Two minor notes, non-blocking:
1. The old `[persist] could not create dir …` warn log is gone from both paths — a failing mkdir now surfaces only via `recordFailure` after the window (or immediately for non-transient codes). Small loss of mid-flight diagnostic; acceptable since the final error is still logged.
2. A *permanent* ENOTDIR (blocker file on an ancestor path) now burns the full ~1.5s window before spilling where it used to fail fast; in `flushSync` that briefly blocks the calling thread. Already called out in the PR body as acceptable — agreed, especially since downstream failure-path tests use small retry configs.

**Tests** (+4, `tests/persist.test.ts`, all mock-injected via `t.mock.method` with 1–2ms retry configs, matching suite convention):
- one-shot rename ENOENT → healed, canonical written, **no** spill asserted;
- one-shot writeFile ENOENT → dir recreated on retry, canonical written;
- persistent ENOENT → throws AND spills, payload survives (mirrors the existing EPERM spill test at L491);
- flushSync variant of the heal case.
One optional gap: no explicit test asserting a non-transient code fails fast without burning the window — the predicate is trivially inspectable, so I'm not requiring it.

**Pre-flight (re-run locally on the PR head, real Node 22):**
- `npm run typecheck` ✓
- `npm test` 575/575 ✓ (matches the claim)
- `npm run build` ✓

**Verdict:** LGTM — this counts as one of the two required independent agent reviews. Merging stays human-side.

PR #182, opened 2026-09-01
fix(panel): correct session-accounting label, mark tokensCompressed cumulative

ranxianglei:
Fixes the panel copy items from ranxianglei/billion-context-pi#267.

## Session-accounting label: "never shrinks" was false

The `Context (session accounting, host footer scale)` line claimed the number "never shrinks; includes compressed originals". Measured in a real session, the host-reported value really does drop after a compress (summaries replace the pruned originals in the window) — e.g. a T2 distillation took it from 55.5k to 48.4k.

Relabeled to `includes compressed originals; shrinks slower than the sent view`: it still carries the compressed originals, so it lags the sent view, but it is not append-only. Doc comments on `StatusPanelInput.tokenCount` and `buildStatusPanel` updated to match.

## Blocks figure marked cumulative

`Blocks: N active / M total (X tokens compressed)` — `state.stats.tokensCompressed` is the cumulative historical total, which read as a contradiction next to the current `Session-only` number. Now labeled `tokens compressed, cumulative`.

## Tests

`tests/panel.test.ts` assertions updated for both labels. Full suite: 568 pass.

Note for the bcp side: after this lands and is published to npm, billion-context-pi should bump its pinned `acp-kernel` to the new version (see ranxianglei/billion-context-pi#268).

PR #210, opened 2026-09-07
release v0.0.56

ranxianglei:
Ships #182 (panel label fixes: never-shrinks wording, cumulative tokensCompressed).

Contents since v0.0.55:
- #182 — panel.ts/sync.ts/types.ts: label accuracy fixes only, no behavior change

After merge, release.yml publishes 0.0.56 to npm automatically. Then billion-context-pi #268 (already rebased onto v0.1.58, CI green) can merge; a follow-up bcp PR will bump the kernel dep to 0.0.56.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "acp-kernel",
-    "version": "0.0.55",
+    "version": "0.0.56",
     "description": "Framework-agnostic context-compression engine (model-driven, 3-tier LSM). Pure core: no host dependency.",
     "license": "MIT",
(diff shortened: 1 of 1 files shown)


ranxianglei:
Correction to the release note above: #182 also included a behavior fix beyond the label changes — syncBlocks now keeps host-expanded (user-decompressed) blocks inactive so restored messages are not re-folded next turn (double cost + lost originals); a fresh compress of the same range creates a NEW block. Full details in billion-context-pi#318 which bumps the dep.

PR #216, opened 2026-09-07
feat(handoff): support persisted folded snapshots (folded + blocksFull)

ranxianglei:
## What

`renderHandoff` grows two optional inputs so hosts that persist a bounded folded snapshot (billion-context #401-style tails) can delegate handoff rendering fully:

- **`folded?: boolean`** — `coreMessages` is an already-pruned persisted snapshot: render it as-is with a `## Conversation (persisted folded snapshot, N messages)` header instead of re-running `prune()`. Re-pruning is wrong there: the snapshot ids no longer align with the state ranges, so `prune()` resurrects dropped summaries at index 0.
- **`blocksFull?: HandoffBlockFull[]`** — per-active-block original content (`{blockId, topic?, count, fullText}`), appended after the conversation when `full && folded`. The folded tail no longer carries the folded ranges' originals; `--full` recovers them from the block content cache.

Backward compatible: both inputs optional, default behavior unchanged (583/583 tests pass, +3 new).

## Why

billion-context PR#434 (issue #601) wants to delete its duplicated handoff renderer and delegate to the kernel; its own renderer learned these two behaviors in #587. This PR gives the kernel parity so the delegation is byte-compatible.

## Consumers

- billion-context `src/export.ts` `renderHandoff(s, full)` — will pass `folded: s.lastMessagesFolded === true` + `blocksFull` built from its `blockContents` cache.

### DIFF
--- a/src/handoff.ts
+++ b/src/handoff.ts
@@ -9,10 +9,26 @@ export interface HandoffMeta {
   extraBullets?: string[];
 }
 
+export interface HandoffBlockFull {
+  blockId: string;
+  topic?: string;
+  count: number;
+  fullText: string;
+}
+
 export interface HandoffInput {
   coreMessages: CoreMessage[];
   state: CompressionState;
   full: boolean;
+  /** coreMessages is an already-pruned persisted snapshot (#401 bounded
+   *  folded tail): render it as-is instead of re-running prune() — its
+   *  message ids no longer align with the state ranges, so pruning would
+   *  resurrect dropped summaries at index 0. */
+  folded?: boolean;
+  /** Original messages per active block, recovered from the block content
+   *  cache. Appended after the conversation when full && folded — the
+   *  folded snapshot itself no longer carries the folded ranges' originals. */
+  blocksFull?: HandoffBlockFull[];
   meta: HandoffMeta;
 }
 
@@ -48,10 +64,13 @@ export function renderHandoff(input: HandoffInput): string {
   if (meta.contextTokens) lines.push(`- last context tokens: ~${meta.contextTokens}`);
   lines.push(`- compression blocks: ${state.blocks.length} (active ${state.blocks.filter((b) => b.active).length})`);
   lines.push("");
-  const view = full ? coreMessages : prune(coreMessages, state);
-  lines.push(full
+  const folded = input.folded === true;
+  const view = full || folded ? coreMessages : prune(coreMessages, state);
+  lines.push(full && !folded
     ? `## Full conversation (${coreMessages.length} messages)`
-    : `## Conversation (folded view as the model saw it, ${coreMessages.length} client messages)`);
+    : folded
+      ? `## Conversation (persisted folded snapshot, ${coreMessages.length} messages)`
+      : `## Conversation (folded view as the model saw it, ${coreMessages.length} client messages)`);
   lines.push("");
   if (view.length === 0) {
     lines.push("No conversation messages to export.");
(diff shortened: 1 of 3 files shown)


ranxianglei:

[bot] 🏷 On it — reviewing PR https://github.com/ranxianglei/acp-kernel/pull/216 now. Pulled the head locally; running typecheck/test/build on the branch and checking the diff against the spec. One thing I already see that I'll flag in my full review: the branch carries a `release v0.0.55` version-bump commit below the feature commit, which the release workflow forbids on non-release branches — I'll detail and fix shortly.

ranxianglei:

[bot] 🏷 Review complete — **approved by two independent reviewers, ready to merge** (merge itself is human-only per repo policy: https://github.com/ranxianglei/acp-kernel/pull/216).

**Correction to my earlier comment:** I flagged a bundled `release v0.0.55` commit on the branch — that was **wrong**. It was an artifact of a stale local master ref in my sandbox clone (its fetch refspec wasn't updating `origin/master`, frozen at `04bd5ed`). True remote master is `44eece88` (= tag v0.0.56, package.json 0.0.56), which already contains `ccbde32` ("release v0.0.55") as an ancestor, and npm's published 0.0.56 matches master exactly — no version drift exists. The branch was simply based on pre-v0.0.56 master. Apologies for the noise.

**What I verified**
- Spec conformance: `folded?: boolean` and `blocksFull?: HandoffBlockFull[]` implemented exactly as described in `src/handoff.ts`; header strings match the spec byte-for-byte; append gated on `full && folded`; `HandoffBlockFull` exported from the barrel next to `HandoffInput`/`HandoffMeta`.
- The core rationale holds up in code: `src/prune.ts` `collectSummaryAnchors` falls back to `insertAt: earliest ?? 0` when a block's raw ids *and* its rendered summary anchor are both absent from the input — so re-pruning a #401-style persisted tail does resurrect dropped summaries at index 0. An explicit `folded` flag is the right layer (making `prune()` idempotent would silently change semantics for every caller passing partial views; `prune`'s existing `injectSummaries` option can't express "input already pruned").
- Backward compatibility: behavior matrix checked for all 4 `{full, folded}` combos — both legacy rows are byte-identical to old behavior; truthy non-boolean `folded` safely falls back via `=== true`; empty/undefined `blocksFull` is a no-op.
- Checks on the final branch: `tsc --noEmit` clean, **586/586 tests pass**, `tsup` build clean. (Your "583/583" was correct for the base you wrote against; current master carries 2 extra sync-config tests from v0.0.56, plus the new tests below.)

**Second independent review:** APPROVE-WITH-NITS, no blockers. Findings addressed in a follow-up commit:
1. Added a contrast test proving the hazard this feature exists for: the same snapshot rendered *without* `folded` resurrects `[Compressed conversation section]` at index 0 — pins the regression guard so a future "simplification" can't delete the branch unnoticed.
2. Extended the full+folded test: topic-less block (no em-dash heading), trailing-newline `fullText` trimming, and blocks-after-conversation ordering.
3. (Informational, no fix) empty-string `fullText` renders an empty section — acceptable; caller data error.

**Branch changes I made** (force-push ×1 + fast-forward):
- Rebased the feature commit onto current master `44eece88` (was based on older master) → `da5f3a4`
- Added the nit tests → `d53b560` `test(handoff): pin folded-snapshot hazard and block-format branches` (+22/−1, tests only)

Final state: 2 commits, 3 files (`src/handoff.ts`, `src/index.ts`, `tests/handoff.test.ts`), mergeable, no conflicts. Ready for you to merge whenever.

PR #229, opened 2026-09-09
fix(search): hybrid max via reduce + LRU doc-cache (#227)

### LINKED ISSUE #227: searchBlocks: hybrid Math.max(...docs) hard-crashes at ~160K docs; docFeatures 8MB cap re-tokenizes large corpora on every call
来源: Tyan66666/billion-context-dsh#133 分析 search_context 长会话劣化（190 块 / 13.3M 遮蔽 token 会话，单次搜索 ~20.8 分钟）时发现

Two independent kernel-layer costs in the search path, both measured on acp-kernel 0.0.29 (the billion-context-dsh host session from #133 + local reproduction at the reported scale).

## 1. `hybridAlgorithm.score` hard-crashes at large doc counts (P1)

`src/search/algorithms/hybrid.ts` normalizes the two channel scores with an argument spread:

```ts
const maxBm = Math.max(...bm.map((r) => r.score), 1e-9)
const maxFz = Math.max(...fz.map((r) => r.score), 1e-9)
```

The spread throws `RangeError: Maximum call stack size exceeded` once the doc count exceeds the JS argument limit (~65K–125K depending on the V8 build).

Reproduced locally with 160,170 docs (46MB text, 190 blocks): the call crashes at `hybrid.ts` before returning any result — the search tool THROWS instead of degrading. The #133 session (19.7K docs) was below the limit, but any session that compresses ~30K+ shadowed messages breaks search entirely.

Suggested fix: compute the max with a loop/reduce (O(n), no argument limit). Two lines.

## 2. `docFeatures` 8MB cap re-tokenizes large corpora on every call

`DEFAULT_CAP_CHARS = 8 * 1024 * 1024`, keyed by full doc text, evicted in insertion order (not LRU). A search corpus larger than the cap re-tokenizes (corpus − cap) on EVERY call:

- Measured (40.4MB corpus, 20,140 docs, CJK-mixed): **17.9s per call, identical across cold/warm calls** — the 8MB window slides and nothing is reused.
- The kernel's own doc-cache docstring estimates ~0.3s/MB for the segmenter pass.
- `setDocCacheCap()` exists and the host now uses it (128MB → 9.0s cold + 0.14–0.16s warm), but the cap design is a multi-session-server tradeoff: steady-state feature heap is ~4–6.3× cached source chars (measured), so a single-session host gets full-corpus caching only at ~800MB+ heap for a 70MB corpus.

Suggested directions (design call for the kernel owner):
- key by doc `ref` with per-session / per-doc-set cache lifetime, so a single-session host can cache the whole corpus without a global char cap;
- or LRU eviction so repeated calls reuse the most-touched docs.

Note: the hybrid double pass (BM25 + fuzzy both scan all docs, then `applyRoleWeight`/`buildResults` each build a full Map) is fine once features are warm (~140ms at 20K docs) — no action needed unless profiling shows otherwise.

## Data

- #133 live session (host 0.1.2-rc.1, plugin v0.2.19): 8 paired tool calls — 6 blocks → 1.8s … 190 blocks / 13.3M shadowed tokens → 1249s; 19.7K docs / ~50MB corpus per call.
- Local reproduction (this report): N=20,861 events / B=190 / 40.4MB corpus: searchBlocks 17.9–18.0s ×3 under the 8MB cap; 9.0s cold + 0.14–0.16s warm under a 128MB cap. 160,170-doc corpus: crash (item 1).

ranxianglei:
Fixes #227.

Two independent kernel-layer costs in the `searchBlocks` path. Both were reproduced locally on the current tree before changing anything.

## Item 1 (P1): `hybridAlgorithm.score` hard-crashes at large doc counts

`src/search/algorithms/hybrid.ts` normalized the two channel scores with an argument spread:

```ts
const maxBm = Math.max(...bm.map((r) => r.score), 1e-9)
```

The spread throws `RangeError: Maximum call stack size exceeded` once the doc count exceeds V8's argument limit (~65K–125K depending on build). **Reproduced**: 160,170 docs → the search tool *throws* instead of degrading, before returning any result.

**Fix**: compute the max with `reduce` (O(n), no argument limit, no extra allocation):

```ts
const maxBm = bm.reduce((m, r) => (r.score > m ? r.score : m), 1e-9)
```

This was the only argument-spread hazard in the codebase (audited every `Math.max/min(...spread)` and array spread).

## Item 2: `docFeatures` 8MB cap re-tokenizes large corpora on every call

The doc-features cache evicted in **insertion order (FIFO)**, keyed by full doc text. A corpus larger than the 8MB cap re-tokenized `(corpus − cap)` on **every** call. **Reproduced**: 12.0MB corpus / 27,741 docs / 8MB cap → cold 7242ms, warm 7524ms (warm/cold = 104%); cache pinned at 8.0MB.

**Fix** (design call — chose LRU over per-scope):
- Eviction is now **LRU**: a cache hit re-inserts the doc at the tail, so the docs a host keeps re-querying are the ones retained (strictly better than FIFO, no interface change).
- `setDocCacheCap(≥corpus)` / `Infinity` is now **documented + tested** as the opt-in for a single-session host to cache the whole corpus once and make later calls `O(docs × query-terms)` — this is what the host needs to stop re-tokenizing, without a per-scope cache redesign.

**Why not per-scope (key by doc `ref`)**: it would require threading a scope through the stateless public `SearchAlgorithm.score(docs, query)` contract — either a breaking change for custom algorithms, or a fragile non-async-safe module-level "active scope". That's a lot of surface for a multi-session-server use case that doesn't exist yet. The single-session host already gets full-corpus caching via the cap knob.

> The hybrid double pass (BM25 + fuzzy both scan all docs, then `applyRoleWeight`/`buildResults` each build a full Map) is left as-is: it's fine once features are warm (~140ms at 20K docs), per the issue.

## Tests

- `hybrid: no RangeError at large doc counts` — 160K docs (the crash scale) returns results instead of throwing.
- `LRU: re-accessing a doc keeps it alive (FIFO would evict it)` — pins the LRU-vs-FIFO difference via object identity.
- `setDocCacheCap(Infinity): whole corpus cached, no eviction` — pins the full-corpus opt-in.
- Renamed the eviction test to "evicts least-recently-used".

Full suite: **583 pass, 0 fail**. Typecheck clean, build succeeds.

### DIFF
--- a/src/search/algorithms/hybrid.ts
+++ b/src/search/algorithms/hybrid.ts
@@ -28,8 +28,9 @@ export const hybridAlgorithm: SearchAlgorithm = {
     score(docs: SearchDoc[], query: string): ScoredBlock[] {
         const bm = bm25Algorithm.score(docs, query);
         const fz = fuzzyAlgorithm.score(docs, query);
-        const maxBm = Math.max(...bm.map((r) => r.score), 1e-9);
-        const maxFz = Math.max(...fz.map((r) => r.score), 1e-9);
+        // reduce, not Math.max(...spread): the spread trips V8's argument limit (~65K–125K docs) → RangeError.
+        const maxBm = bm.reduce((m, r) => (r.score > m ? r.score : m), 1e-9);
+        const maxFz = fz.reduce((m, r) => (r.score > m ? r.score : m), 1e-9);
         const bmMap = new Map(bm.map((r) => [r.ref, r.score / maxBm]));
         const fzMap = new Map(fz.map((r) => [r.ref, r.score / maxFz]));
         return docs.map((d) => ({
--- a/src/search/doc-cache.ts
+++ b/src/search/doc-cache.ts
@@ -10,10 +10,18 @@
  * processed once; later searches are O(docs × query-terms).
  *
  * Keyed by doc text (immutable). Bounded by total cached source chars —
- * oldest docs are evicted when the cap is exceeded, so a long-lived
- * process serving many sessions cannot grow unboundedly. Hosts that want to
- * release the memory eagerly on session shutdown/switch can call
- * clearDocFeatures() (optional: the cap already bounds it).
+ * least-recently-used docs are evicted when the cap is exceeded (LRU), so a
+ * long-lived process serving many sessions cannot grow unboundedly, and the
+ * docs a host keeps re-querying are the ones retained.
+ *
+ * The default cap (8MB) is a conservative bound for multi-session servers.
+ * A single-session host whose corpus exceeds it would otherwise re-tokenize
+ * (corpus − cap) on EVERY call; call setDocCacheCap() with a value ≥ corpus
+ * size (or Infinity) to cache the whole corpus once and make later calls
+ * O(docs × query-terms).
+ *
+ * Hosts that want to release the memory eagerly on session shutdown/switch
+ * can call clearDocFeatures() (optional: the cap already bounds it).
  */
 
 import { charBigrams, tfMap } from "./tokenizer.js";
@@ -44,7 +52,12 @@ function build(text: string): DocFeatures {
 
 export function docFeatures(text: string): DocFeatures {
     const hit = cache.get(text);
-    if (hit) return hit;
+    if (hit) {
+        // LRU: re-insert at the tail (most-recently-used); eviction takes from the head.
+        cache.delete(text);
+        cache.set(text, hit);
+        return hit;
+    }
     const f = build(text);
     if (text.length > 0 && text.length <= capChars) {
         while (cachedChars + text.length > capChars && cache.size > 0) {
@@ -66,7 +79,9 @@ export function clearDocFeatures(): void {
 
 /**
  * Set the cache cap in source chars. Docs larger than the cap are never
- * cached. Also used by tests to exercise eviction.
+ * cached. Pass a value ≥ the corpus size (or Infinity) to cache the whole
+ * corpus — a single-session host then avoids re-tokenizing on every call.
+ * Also used by tests to exercise eviction.
  */
 export function setDocCacheCap(chars: number): void {
     capChars = Math.max(1, chars);
--- a/src/search/doc-cache.ts
+++ b/src/search/doc-cache.ts
@@ -4,21 +4,35 @@
  * A search over the compressed history re-scores the SAME immutable docs on
  * every call — compressed block summaries and folded message text never
  * change. Without this cache, every search_context call re-tokenized the
- * entire corpus (segmenter CJK pass ≈ 0.3s/MB cold) plus re-lowercased it
- * and rebuilt the bigram set for each channel: a 5MB session cost ~3s PER
- * CALL, growing linearly with session length. With the cache the corpus is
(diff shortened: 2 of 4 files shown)


ranxianglei:

Reviewed PR #229 against the live tree — reproduced both bugs pre-change, then re-ran the full suite on the PR branch. **Verdict: mergeable.** Both fixes are correct; 583/583 tests pass, typecheck clean, build green. Two notes below: a framing clarification on P2, and a minor finding the "only hazard" claim misses.

**P1 — hybrid `Math.max(...spread)` → `reduce`: verified.**
- Pre-PR tree, 160K docs → `RangeError: Maximum call stack size exceeded` (search throws, returns nothing). Post-PR: returns 10 correctly-scored results in ~1.2s.
- The `reduce` is semantically identical to the old `Math.max(...scores, 1e-9)`, including the empty-array → `1e-9` floor. No behavior change at small N.
- The new 160K-doc test is a valid crash-scale regression test (throws on old, passes on new).

**P2 — LRU + cap knob: correct, but one framing point.**
- LRU implementation is sound (hit → re-insert at tail; eviction takes the head); strictly ≥ FIFO.
- **But LRU alone does not fix the reported symptom.** I measured a 12MB corpus under the 8MB default cap: old FIFO warm/cold ≈ 132%, new LRU warm/cold ≈ 137% — no improvement. The host re-scores the *same* corpus in the *same* order every call, so with cap < corpus the 8MB window slides and every doc is a miss each pass, FIFO or LRU.
- **The actual fix is the cap knob** (`setDocCacheCap(≥corpus)` / `Infinity`): I measured warm/cold ≈ 14% (7× faster), whole corpus cached once. That's what the host needs, and the PR correctly documents + tests it.
- Net: the host must still set the cap ≥ corpus to get the win. LRU is a genuine robustness improvement for varied/subset access, but the "chose LRU" framing could lead a host to expect the upgrade alone to stop re-tokenization. Suggest one line in the host-facing note: "set the cap; LRU does not by itself stop same-order full-scan re-tokenization."

**Minor — the "only argument-spread hazard" claim is inaccurate.**
- `src/search/tokenizer.ts:102`: `tokens.push(...cjkRunTokens(segs))` is the same class (spread into a function call). Verified `push(...)` throws `RangeError` at 150K args on this Node build (same as `Math.max`); array-literal spread `[...]` is safe.
- Trigger: a single doc with a very long contiguous CJK run — all-OOV run >~75KB (`cjkRunTokens` returns ~2L tokens) or a dictionary run >~300KB. Far rarer than the hybrid case (which only needed 65K+ small docs), so lower severity, but a real latent crash in the search path.
- Trivial follow-up: replace the `push(...)` with a `for...of`. The other function-call spreads (`src/decompress.ts:74/82`, `src/compress.ts:347/361`) are the same class but bounded by block/warning counts — very unlikely to trip, worth a glance in the same pass.

**Merge cleanliness:** PR is based on `04bd5ed`; master is 32 commits ahead but none touch the 4 changed files, so it should merge cleanly (dry-run merge: no conflicts).

Happy to open a follow-up issue for the tokenizer `push(...)` if you want it tracked.

PR #226, opened 2026-09-09
fix: break the reasoning context floor and live-anchor summary duplication

ranxianglei:
Fixes #224

## Problem

Measured on a real storm session (billion-context-paper, qwen3.8-27b), ~50% of visible context was replayed thinking pinned to compress-anchor messages, and live compress-call args duplicated the rendered `acp_summary` text. Five root causes were verified in code (see #224 for the full measurement).

## Changes

1. **`Config.reasoningReplay`** (`"always" | "open-round" | "never"`, default `"always"` = legacy). New `strip-reasoning` pipeline node (after `sync-blocks`, before `prune`) drops closed-round reasoning by round-closure gating on the last real user message. Provider replay only requires the *open* round's thinking (Anthropic signature / Gemini thought_signature), so the gate is provider-agnostic. `"always"` is a no-op kill-switch.

2. **Base-id coverage normalization.** `prune`/`sync`/`recommend`/`dangling` coverage now normalizes on the base id (`baseMessageId`), so a host `#suffix` projection (`base#r0`) is covered whenever any sibling projection is — stops historical messages resurrecting as reasoning cores after projection evolution (the 762K resurrection under `always`).

3. **Live compress-call args.** `rewriteCompressText` now skips the ref-tag prefix before `JSON.parse` (previously it always failed on tagged calls, so the rewrite never applied) and stubs the summary to a 200-char prefix, since the rendered `acp_summary` is the authoritative copy.

4. **`ProcessTurnInput.extraTokens`.** Host-reported kernel-invisible mass is folded into usage/pressure so the pressure band tracks real provider-side usage instead of only the kernel-visible view.

## Measured impact (real storm session replay)

- `open-round`: visible ~155K → **48.8K** (thinking 75.8K → 0.9K, only the open round kept)
- compress-call args: 21.6K → 16.5K (stubbing effective)
- `always` (kill-switch): behavior identical to before

## Verification

- `npm run typecheck` — clean
- `npm test` — 599/599 pass (19 new tests in `tests/reasoning-replay.test.ts`)
- `npm run build` — success

Issue #231, opened 2026-09-09
prune: isCovered misses base-id coverage when block effective ids use sub-id forms (base#suffix)


ranxianglei:
Found while investigating billion-context-pi #336 (closed PR #225).

**Bug:** `isCovered(id, covered)` in src/prune.ts only tests exact id membership. If a compression block's recorded effective message ids use sub-id forms (`base#callId`, `base#r0`, …) while the current projection emits plain `base` (or vice versa after a projection change), coverage checks fail and already-compressed messages re-enter the visible view resurrected (measured: a 762K-token leak in replay after introducing thinking sub-id projections).

**Fix:** normalize to the base id before membership — `covered.has(id) || covered.has(base) || any covered entry shares the same base prefix`. Currently latent (no shipped adapter projects sub-ids), but it bites anyone who extends the projection.

ranxianglei:
> 🤖 Powered by [ework](https://github.com/ranxianglei/ework-aio) · qwen3.8-27b

[bot] 🏷 Triage update — bug verified, fix in progress.

**Verified (repro, not just code reading):** added 8 regression tests; 6 fail on current master:
- `prune` resurrects the message in **both** directions: block records sub-ids + view emits plain base (your primary scenario), and block records base + view emits sub-ids.
- Sibling sub-ids of the same base (`h_abc#r0` covered, view emits `h_abc#call1`) also leak.
- Negative case (distinct id `h_abcde` vs covered base `h_abc`) correctly stays visible — the fix must not turn into prefix matching.

**Mechanism:** `src/prune.ts:144` does `covered.has(messages[index]!.id)` — exact string membership, with `covered` built verbatim from `block.effectiveMessageIds` (`src/state.ts:45`). Root cause is exact-string coverage in a world where one original message can appear under several id forms (`base`, `base#r0`, `base#callId`); the 762K leak is the symptom. Your layering is right.

**Two findings beyond the issue text:**
1. The `isCovered` from closed PR #225 early-returns when the id has no `#` (`hash <= 0`), so it **misses your primary direction** (block recorded sub-ids, view emits plain base). My fix normalizes *both* sides to the base, so all three combinations are covered, with an O(1) precomputed base set instead of the per-message linear scan.
[...]

PR #232, opened 2026-09-09
fix(hide-consumed): parse tag-prefixed compress call text, stub live anchor summaries

### LINKED ISSUE #230: hide-consumed: parseCallText fails on tag-prefixed compress call text, so live anchor args are never stubbed
Found while investigating billion-context-pi #336 (PR #225, closed).

**Bug:** `rewriteCompressText`/`parseCallText` in src/hide-consumed.ts JSON.parse the compress tool-call text starting at index 0. When the adapter message carries a leading metadata tag (pi appends `<acp ...>mNNNNN</acp>` tags to message text), the parse fails, the call is treated as unparseable, and **live compress anchors keep their full summary args forever** — the summary text is duplicated in every request (measured 21.6K tokens of duplicated args in session 01a07b3c).

**Fix:** locate the JSON by scanning for the first `{` before parsing (as done in the closed PR #225). Independent of any reasoning feature — worth shipping on its own.

ranxianglei:
Fixes #230.

## Bug

`rewriteCompressText` in `src/hide-consumed.ts` JSON.parsed the compress tool-call text starting at index 0. Adapters (pi) persist the rendered ref tag in front of the tool-call text, so the parse always threw on tagged calls and the call was treated as unparseable:

1. consumed sibling entries were never dropped from kept calls' args;
2. live compress anchors kept their full summary text in the args forever — duplicating the rendered `acp_summary` message in every request (~22K tokens of duplicated args measured in session 01a07b3c, billion-context-pi #336).

Reproduced on master before the fix: a tag-prefixed mixed call (live + consumed entries) kept both entries, and a fully-live call with a 5000-char summary kept all 5000 chars in its args.

## Fix

- `parseCallText`: locate the JSON by scanning for the first `{` before parsing (the closed PR #225 approach), preserving the tag prefix on output.
- Stub entry summaries over 200 chars in kept compress calls (active and newest orphans); the rendered block summary remains the authoritative copy.

Independent of any reasoning feature — shipped on its own per #230.

## Relationship to PR #226

PR #226 (open) bundles the same fix as its item 3, using a regex that strips the exact ref-tag prefix and without stubbing orphaned calls. This PR implements the `{`-scan approach prescribed in #230 (handles any leading metadata; also stubs orphans). If this PR merges first, #226 should drop its hide-consumed hunk; if the regex variant is preferred, this PR can be reworked.

## Verification

- `npm run typecheck` — clean
- `npm test` — 628/628 pass (4 new tests in `tests/hide-consumed-stub.test.ts`)
- `npm run build` — success

### DIFF
--- a/src/hide-consumed.ts
+++ b/src/hide-consumed.ts
@@ -22,7 +22,7 @@ function rangeKey(startRef: string, endRef: string): string {
 // Adapters (pi) persist the rendered ref tag in front of the tool-call text,
 // so the JSON args no longer start at index 0. Locate the first "{" instead of
 // parsing the raw text — the prefix is preserved on output.
-function parseCallText(text: string | undefined): { prefix: string; obj: Record<string, unknown>; content: unknown[] } | null {
+function parseCallText(text: string | undefined): { prefix: string; obj: Record<string, unknown>; content: unknown[]; contentWasString: boolean } | null {
     const raw = text ?? "";
     const start = raw.indexOf("{");
     if (start < 0) return null;
@@ -34,14 +34,31 @@ function parseCallText(text: string | undefined): { prefix: string; obj: Record<
     }
     if (!parsed || typeof parsed !== "object") return null;
     const obj = parsed as Record<string, unknown>;
-    if (!Array.isArray(obj.content) || obj.content.length === 0) return null;
-    return { prefix: raw.slice(0, start), obj, content: obj.content };
+    let content: unknown[] | null = null;
+    let contentWasString = false;
+    if (Array.isArray(obj.content)) {
+        content = obj.content;
+    } else if (typeof obj.content === "string") {
+        // Non-strict-tool providers (qwen etc.) sometimes stringify the content
+        // array inside the JSON args; the compress tool accepts it, so the
+        // rewrite must too. Measured: ALL 52 calls in the billion-context-pi
+        // #336 storm session used this form (#230).
+        contentWasString = true;
+        try {
+            const inner: unknown = JSON.parse(obj.content);
+            if (Array.isArray(inner)) content = inner;
+        } catch {
+            content = null;
+        }
+    }
+    if (!content || content.length === 0) return null;
+    return { prefix: raw.slice(0, start), obj, content, contentWasString };
 }
 
 function rewriteCompressText(text: string | undefined, liveKeys: Set<string>): string | null {
     const parsed = parseCallText(text);
     if (!parsed) return null;
-    const { prefix, obj, content } = parsed;
+    const { prefix, obj, content, contentWasString } = parsed;
 
     const kept = content.filter((entry): entry is Record<string, unknown> => {
         if (!entry || typeof entry !== "object") return false;
@@ -53,7 +70,7 @@ function rewriteCompressText(text: string | undefined, liveKeys: Set<string>): s
 
     if (kept.length === 0) return null;
 
-    return prefix + serializeCompacted(obj, kept).text;
+    return prefix + serializeCompacted(obj, kept, contentWasString).text;
 }
 
 // Live compress-call args duplicate every range's full summary text while the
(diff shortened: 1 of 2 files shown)


ranxianglei:
Linking: fixes #230 (tag-prefixed call text parse failure). Verified locally on 8a9ffbd: 628/628 tests, typecheck + build clean. Note: the separate latent prune sub-id coverage bug is tracked in #231 (no PR yet — no shipped adapter projects sub-ids).

ranxianglei:

Independent review on `8a9ffbd` — verified and looks good:

**Verification (fresh checkout, not your local run)**
- `npm run typecheck` — clean
- `npm test` — 628/628 pass (7 suites)
- `npm run build` — success

**Code review**
- `parseCallText` first-`{` scan (`src/hide-consumed.ts:25-39`): safe — the ref-tag prefix (`<acp tokens="…" type="…">ref</acp>\n`, `src/render-refs.ts:39-46`) contains no braces, and scanning for `{` is strictly more general than #226's exact-tag regex (survives any future leading metadata).
- `rewriteCompressText` behavior change: fully-live calls are now re-serialized too (old code bailed when `kept.length === content.length`). Output is idempotent and byte-stable across turns for compact JSON, so no churn.
- Stubbing: `slice(0, 199) + "…"` → exactly 200 chars, only applied when `summary.length > 200`; safe because the rendered `acp_summary` message remains the authoritative copy. Orphan stubbing via the `compactCompressText` fallback is correctly scoped to the newest orphans (`KEEP_LAST_ORPHANED`).
- Pipeline ordering is fine: `hide-compress-calls` runs before `render-refs` (`src/compress.ts:572`); the preserved prefix is stripped and re-rendered with fresh token accounting by `render-refs`, so no stale tags leak.
- Tests: all 4 new cases are meaningful — including building the tagged input from real `renderVisibleRefs` output rather than a hand-written prefix, and the mixed-call case asserting the consumed sibling is dropped with the live entry byte-identical. No `as any` / `@ts-ignore`.

**Process notes**
- This counts as the second independent agent review (authoring session + this one) per AGENTS.md §6.
- #226 overlap: its item 3 is the same fix via the regex variant. If this PR merges first, #226 should drop its hide-consumed hunk — your call.
- #231 (prune sub-id coverage) correctly split out, no PR needed until an adapter projects sub-ids.

Ready to merge on your side — I can't merge PRs (human-only).

ranxianglei:
**Extended:** forensic replay of the #336 storm session showed ALL 52 compress calls use `content` as a **stringified array** (non-strict-tool provider form), 0 tag-prefixed — so the index-0 scan alone would still not have compacted that session. Added: parseCallText accepts the string form; serializeCompacted re-stringifies to preserve the original shape. Verified on the real session's last call: 924 → 278 chars, string shape kept, live range kept. 629/629 tests.

ranxianglei:

Reviewed the extension on `cbc78fb` — verified and good to go:

**Verification (fresh checkout)**
- `npm run typecheck` — clean
- `npm test` — 629/629 pass
- `npm run build` — success

**Diff review**
- `parseCallText` string path (`src/hide-consumed.ts:37-52`): safe on all edge shapes — inner `JSON.parse` throwing or yielding a non-array (numbers, `null`, objects, garbage) leaves `content = null` → no rewrite; empty inner array bails like the array path.
- `serializeCompacted` shape preservation: `JSON.stringify(compacted)` embedded as a JSON string keeps `{"content":"[...]"` byte-shaped; key order is preserved since `content` is already a key on the spread `obj`. Round-trip is idempotent — re-parsing a rewritten call yields stubbed (≤200-char) summaries, so `compactCompressText` reports no change and the text stops churning.
- New test asserts the full round-trip: string shape preserved, consumed sibling dropped, live entry stubbed to exactly 200 chars ending in `…`.

**On the forensic finding** — this is the important one, and it changes which fix actually addresses #336: on master, a stringified-content call parses fine at index 0 but then bails at the `Array.isArray(obj.content)` check, so the tag-prefix fix alone (8a9ffbd) would indeed not have compacted that session. The #230 repro (tag prefix) and the #336 storm (stringified content) were two distinct parse-failure modes of the same fragile parser; both are now handled. The 924 → 278 char number comes from your replay — I verified the mechanism via the test path rather than the session data, but the stub math is consistent.

Ready to merge on your side — I can't merge PRs (human-only).

PR #250, opened 2026-09-11
feat(nudge): option to render nudges without the guidance sections

### LINKED ISSUE #248: feat(nudge): option to render nudges without the guidance sections
## Problem

`renderNudgeText(decision, prompts?)` always embeds all four guidance texts — `COMPRESS_PHILOSOPHY`, `HOW_TO_COMPRESS_RULES`, `TIER2_DISTILL_RULES`, `TIER3_CONDENSE_RULES` (from `compression-rules.ts`) — into the nudge body.

A host that already delivers these same four texts **once** via the system prompt and tool descriptions re-bills ~6 KB of identical text on **every** nudge injection. Measured in a long live session (billion-context-dsh PR #143, B6 "slim nudges"): nudge body ≈ 6.1 KB × 3 injections ≈ 18.4 KB of repeated tokens per stretch of turns, zero new information.

## Why the existing override path cannot express this

`resolvePrompts()` is the only knob today, but the resulting `Prompts` object feeds **both** the nudge renderer and the system-prompt composition (the docs say exactly that: resolve once at startup, pass to `renderNudgeText` *and* to the adapter's system-prompt composition). Overriding a field therefore changes it on **both** surfaces at once. The host needs the opposite split: keep the guidance in the system prompt (read once), drop it from every nudge (re-read each injection). There is currently no way to say "system prompt: yes, nudge: no".

## Proposed fix

An opt-out at nudge-render time, e.g.:

```ts
renderNudgeText(decision: NudgeDecision, prompts?: Prompts, options?: { includeGuidance?: boolean }): RenderedNudge
// includeGuidance defaults to true → fully backwards compatible
```

or equivalently, return the rendered nudge as structured parts so a host can drop the guidance part when composing. The system-prompt composition path should be unaffected by the option.

## Consumer + current workaround

billion-context-dsh (DeepSeek Harness ACP engine, https://github.com/Tyan66666/billion-context-dsh/pull/143) pins acp-kernel 0.0.29 and post-processes the rendered text today: `stripNudgeGuidance` (src/nudge.ts) removes the four exported blocks by exact string match after rendering. It is a labeled host-side workaround with a removal gate in its docs (`docs/dsh-porting-verification.md`): delete it and use this option once it exists. Exact-match stripping is fragile — any rewording of a guidance section silently disables the host-side strip — which is why the clean fix belongs here.

Found while implementing B6 ("slim nudges") in billion-context-dsh PR #143.

ranxianglei:
## Problem

`renderNudgeText` always embeds all four load-bearing guidance texts (`COMPRESS_PHILOSOPHY`, `HOW_TO_COMPRESS_RULES`, `TIER2_DISTILL_RULES`, `TIER3_CONDENSE_RULES`) into every nudge body. A host that already delivers these same texts once via the system prompt + tool descriptions re-bills ~6 KB of identical text on **every** injection (~18.4 KB per stretch of turns in the B6 measurement), zero new information.

The existing override path (`resolvePrompts`) cannot express this split: its result feeds both the nudge renderer and system-prompt composition, so overriding a field changes both surfaces at once.

## Fix

Opt-out at nudge-render time, exactly as proposed in #248:

```ts
renderNudgeText(decision: NudgeDecision, prompts?: Prompts, options?: { includeGuidance?: boolean }): RenderedNudge
// includeGuidance defaults to true → fully backwards compatible
```

With `includeGuidance: false`, exactly the four exported guidance blocks are dropped (philosophy from both the efficiency note and emergency header, HOW TO COMPRESS, and the active tier's distill/condense rules). All per-injection dynamic content survives: context breakdown, trigger line, compressible ranges, tier target block list, the JSON call-shape example, and the batch-call hint — with no leftover stacked blank lines.

Chosen over the structured-parts alternative: additive, keeps `RenderedNudge.text` stable for existing hosts; structured parts would widen the public surface for what is effectively one boolean.

## Changes

- `src/nudge-text.ts` — new exported `RenderNudgeOptions`; all three render branches (gentle / emergency / tier) conditionally append the guidance sections
- `src/index.ts` — export `RenderNudgeOptions`
- `src/prompts.ts` — `resolvePrompts` JSDoc now cross-references the option so hosts don't try to express the surface split via field overrides
- `tests/nudge-text.test.ts` — 7 new tests: backwards-compat identity across all 4 modes, slim output per mode (guidance absent, chrome present, no stacked blanks), default tier nudges still embed only the active tier's rules, size-savings sanity (< 50% of full text); fixture now carries a realistic `contextBreakdown` (decideNudge always populates it)

## Verification

- `npm run typecheck` — clean
- `npm test` — 651/651 pass
- `npm run build` — success

## Consumer

billion-context-dsh (B6 "slim nudges", https://github.com/Tyan66666/billion-context-dsh/pull/143) pins acp-kernel 0.0.29 and post-processes rendered text today (`stripNudgeGuidance`, exact-match stripping of the four blocks — fragile against any rewording). Once this ships, it can delete that workaround and pass `{ includeGuidance: false }`.

Closes #248

### DIFF
--- a/DESIGN-prompts.md
+++ b/DESIGN-prompts.md
@@ -72,6 +72,12 @@ export function resolvePrompts(
 - `renderNudgeText(decision, prompts = defaultPrompts)` now accepts a `Prompts`
   argument, so an override flows into every nudge tier (gentle / emergency /
   tier-2 / tier-3). The default keeps the one-arg call backward-compatible.
+- The renderer also accepts `options?: { includeGuidance?: boolean }` (default
+  `true`; exported as `RenderNudgeOptions`). Set `false` when the host already
+  delivers the four rules once via system prompt / tool descriptions: exactly
+  those texts are dropped from the nudge body, all per-injection dynamic content
+  survives, and the system-prompt surface is untouched. Field overrides cannot
+  express this split — they change both surfaces at once.
 
 ### Host usage
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -51,7 +51,7 @@ export {
 } from "./tokenize.js";
 export type { TokenCountFn } from "./tokenize.js";
 export { renderNudgeText, formatRanges } from "./nudge-text.js";
-export type { NudgeVoice, RenderedNudge } from "./nudge-text.js";
+export type { NudgeVoice, RenderedNudge, RenderNudgeOptions } from "./nudge-text.js";
 export {
   COMPRESS_PHILOSOPHY,
   HOW_TO_COMPRESS_RULES,
--- a/src/nudge-text.ts
+++ b/src/nudge-text.ts
@@ -9,12 +9,30 @@ export interface RenderedNudge {
   text: string;
 }
 
-function efficiencyNote(prompts: Prompts): string {
-  return `This is an efficiency nudge to compress early and keep context lean — not an overflow warning. A separate, stronger alert will appear if the context is actually full.\n\n${prompts.compressPhilosophy}`;
+export interface RenderNudgeOptions {
+  /**
+   * Embed the load-bearing guidance texts (compressPhilosophy, howToCompressRules,
+   * and the active tier's distill/condense rules) in the nudge body. Defaults to
+   * true, so existing callers render byte-identical output. Set to false when the
+   * host already delivers these same texts once via its system prompt or tool
+   * descriptions — re-billing them on every injection is pure repetition. Only
+   * the four guidance texts are dropped; per-injection dynamic content
+   * (breakdown, trigger line, ranges, tier target blocks) survives. The
+   * system-prompt composition path reads {@link Prompts} directly and is
+   * unaffected by this flag — this is how a host keeps guidance in the system
+   * prompt while slimming every nudge.
+   */
+  includeGuidance?: boolean;
 }
 
-function emergencyHeader(prompts: Prompts): string {
-  return `⚠️ Context limit reached — compress now. Prioritize consumed tool outputs.\n\n${prompts.compressPhilosophy}`;
+function efficiencyNote(prompts: Prompts, includeGuidance: boolean): string {
+  const note = "This is an efficiency nudge to compress early and keep context lean — not an overflow warning. A separate, stronger alert will appear if the context is actually full.";
+  return includeGuidance ? `${note}\n\n${prompts.compressPhilosophy}` : note;
+}
+
+function emergencyHeader(prompts: Prompts, includeGuidance: boolean): string {
+  const header = "⚠️ Context limit reached — compress now. Prioritize consumed tool outputs.";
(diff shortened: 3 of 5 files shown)


ranxianglei:

[bot] 🏷 **PR review — verdict: mergeable** (merge itself stays human-only).

**Duplicate screening**: this PR is the implementation of #248 (the proposal); not a duplicate — `Closes #248` will auto-close it on merge.

**Diff cleanliness (first gate)**: clean. 1 commit, 4 files, +166/−53 — every line serves the feature; no unrelated files, no lockfile churn, `package.json` untouched (no version bump), no `as any`/`@ts-ignore`.

**Verification** (all run locally at the PR head on real Node v22):
- `npm run typecheck` clean · `npm test` 651/651 pass · `npm run build` success — matches the PR claims.
- Independent byte-identity check beyond the PR's own tests: I rendered all four modes plus edge cases (empty ranges, protected ranges, dangerous flag, missing breakdown, zero growth, over-limit, emergency+tier-2/3) through both the base commit's implementation and the new one — default-mode output is byte-identical in 12/12 cases. Slim mode drops exactly the four guidance blocks in every mode, keeps all chrome (breakdown, trigger line, ranges, target blocks, JSON example, batch hint), leaves no stacked blank lines, and lands at ~7–11% of full-text size (spec required <50%).
- Fixture comment verified true: `decideNudge` unconditionally populates `contextBreakdown` (`src/compress.ts:1400,1430`).
- `RenderNudgeOptions` confirmed exported from the built `dist/index.d.ts`.

**One edge case noted, deliberately NOT fixed**: a hand-crafted decision with `contextBreakdown` undefined renders stacked blank lines in slim mode — but the *old* code produces byte-identical output for that input too, so this is pre-existing behavior, unreachable via `decideNudge`, and "fixing" it would break this PR's explicit byte-identity backwards-compat contract. Out of scope by design.

**Small fix applied directly on the branch** (per review rules, committed as `e6bcd62`): `DESIGN-prompts.md` documented the stale signature `renderNudgeText(decision, prompts = defaultPrompts)` right next to its discussion of the system-prompt/nudge surface split — exactly where a host reader looks for this option. Added a short bullet documenting `{ includeGuidance?: boolean }` / `RenderNudgeOptions` so the design doc doesn't send readers back to field overrides. Docs-only; branch now has 2 commits, still mergeable with zero conflicts (master moved to v0.0.63 since the base, but shares no files with this PR — no rebase needed).

For billion-context-dsh B6: once this lands in a release, `stripNudgeGuidance` can be deleted in favor of `{ includeGuidance: false }`.

中文摘要：审查了 includeGuidance 选项的 PR——默认输出与旧版逐字节一致（独立验证 12 个场景含边界用例）、slim 模式无残留空行且体积降至全文约 10%；发现并直接修复了 DESIGN-prompts.md 中过期的函数签名文档（docs commit e6bcd62），可以合并。

PR #252, opened 2026-09-11
feat: add active-block ledger to nudge text (#251)

ranxianglei:
## Problem (#251)

The nudge text lists recommended compressible ranges but carries no map of existing active blocks. When the model's memory of prior compress coverage drifted from actual coverage (protected-message exclusions, tool-pair boundary widening, consumed-block merges — all of which change `effectiveMessageIds` vs the requested range), it had to spend an `acp_status` round-trip (~1.8K tokens) to reconcile before acting.

## Changes

- `NudgeDecision` gains two optional fields: `activeBlocks` (all active blocks from `activeBlocks(state)`) and `messageRefs`. Populated by `decideNudge`; backward compatible for hosts that build decisions by hand.
- `renderNudgeText` appends a compact ledger line after the ranges section in **all three voices** (gentle / emergency / tier-2/3):
  ```
  Blocks: b1=m00001–m00043 · b2=m00057–m00078 · b3=tier2(m00001–m00078)
  ```
- Span = min/max ref of `effectiveMessageIds` resolved through `byRaw` (skips `BLOCKED` and unresolvable ids). Fallback to the stored spec span only when it parses as m-refs (block-boundary calls store block IDs like `b1` there, which would mislead); otherwise the block is omitted rather than shown with a wrong span.
- Truncation above `MAX_BLOCK_LEDGER_ENTRIES` (8): newest 8 shown + `( +N older)` count. Constant exported for tuning.
- `formatBlockLedger` exported for direct testing; DESIGN.md type sketch updated.

## Verification

- `npm run typecheck` ✅
- `npm test`: 588 pass / 0 fail (8 new tests: normal spans, single-message span, tier annotation incl. tier-nudge rendering, truncation, empty/unresolvable → no line, spec-span fallback + block-ID spec rejection, BLOCKED skipping, emergency placement)
- `npm run build` ✅

Cost ~50–100 tokens per nudge; opencode-acp and billion-context-pi both benefit without host changes (fields are optional).

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -169,6 +169,8 @@ type NudgeDecision = {
     shouldInject: boolean;
     reason: string;
     compressibleRanges: { startRef: string; endRef: string; tokens: number }[];
+    activeBlocks?: CompressionBlock[];   // rendered as a compact block ledger (id → ref span) in the nudge text (#251)
+    messageRefs?: MessageRefMap;         // resolves each block's effectiveMessageIds back to m-refs
     contextUsage: number;       // 0..1
     tier: 1 | 2 | 3 | null;     // multi-tier trigger, if any
     breakdown: Record<string, number>;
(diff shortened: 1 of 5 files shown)


ranxianglei:
Superseded by #253 — both PRs implement #251 (active block map in nudge text), but #253 replaces this draft's approach: spans come from each block's stored `startRef`/`endRef` with a fallback to `effectiveMessageIds`→ref-table, plus a new exported `src/block-map.ts` module (`resolveBlockSpan` / `activeBlockSpans` / `formatCreatedBlocks`) and per-range user-message counts.

Closing to avoid the two open PRs being merged against each other. Reopen if you want to keep it around for comparison.

Issue #211, opened 2026-09-07
parseCompressArgs: fold single-quote JSON repair into the salvage ladder


ranxianglei:
来源: ranxianglei/billion-context#603 (https://github.com/ranxianglei/billion-context/issues/603) 分析 lenient compress-arg parsing 时发现

## Problem

`parseCompressArgs` (as of 0.0.54) already salvages most malformation classes — fence stripping, trailing commas, raw newlines in strings, double-stringified unwrap, truncated/prose-wrapped content-array entry extraction — but **single-quoted JSON** still returns 0 ranges:

```
input: {'content':[{'startId':'m00010','endId':'m00020','summary':'first'}]}
0.0.54 → 0 ranges, kind=malformed-json

input (object form, stringified content): {"content": "[{\'startId\':\'m00010\'}]"}
0.0.54 → 0 ranges, kind=no-valid-ranges
```

Weak local models emit exactly this shape (field evidence: ranxianglei/billion-context-omp#121), and every hit wastes a compress round and hands the model a FAILED result that can trigger tag-echoing.

## Interim mitigation (shipped, host-side)

billion-context PR #610 adds a salvage fallback in the proxy's `parseCompressInput`: when the kernel recovers zero ranges or reports invalid items, retry once through a single→double quote-normalizing state machine (apostrophes inside double-quoted values are data and are copied verbatim; control chars inside single-quoted regions become JSON escapes). The retry wins only if it recovers strictly more ranges, so valid input is never rewritten. Reference implementation: https://github.com/ranxianglei/billion-context/blob/2026-09-07_lenient-compress-args/src/compress-tool.ts (`normalizeSingleQuotes`) — both shapes above recover 1 range with it.

## Suggestion

Fold the same quote-repair into the kernel's salvage ladder (inside/around `tryParseLenient`, before entry extraction) so all hosts (billion-context proxy, omp, pi) converge on one parser, then drop the host fallback after the next kernel release. The state machine is ~60 lines, pure text repair (never invents structure), and its test matrix (full single-quoted object, mixed quotes w/ apostrophe in data, truncated array, prose-with-apostrophes stays rejected, valid input untouched) is in billion-context tests/basic.test.ts (#603 cases).

ranxianglei:
> 🤖 ework agent · qwen3.8-27b

[bot] 🏷 收到，已开工。先说结论:

[...]

PR #269, opened 2026-09-13
feat(compress): line-form content entries — one string per range

### LINKED ISSUE #268: compress tool input fails at non-strict-tool providers: nested JSON mangled in tool arguments
## Problem

At non-strict-tool providers (vLLM, GLM, omp gateways), tool arguments are frequently delivered as **strings with mangled nested JSON** — single quotes, unescaped inner quotes, dropped braces. The compress tool's content schema requires an array of objects \`{startId, endId, summary}\`, so calls from these providers fail at the input boundary (content-not-array / no-valid-ranges) even when the model wrote a correct, detailed summary. Host-side diagnostics (billion-context \`[acp-compress-input] rejected\` logs) show quote-salvage recovering some cases but the summary body itself is often corrupted by then.

Root cause: the object shell around the summary text is pure liability for these providers — every quote and brace inside the markdown must survive JSON escaping twice (model output → stringified args → parser).

## Proposal

Accept a **line form** for content entries: one string per range —

\`\`\`
m00150–m00220 optional topic
## Summary written verbatim
body with pipes | quotes " ... backslashes \\ — all opaque
\`\`\`

First line = refs header (pair or single ref, en/em-dash/~/…/to separators, optional trailing topic); remaining lines = the summary markdown verbatim. Only the refs are structural; the summary body needs no escaping. Legacy object form keeps working; the two mix freely in one call. Headerless entries get a derived topic (first heading / first line, truncated) — display metadata only.

## Acceptance
- schema (anthropic + openai shapes): content items = anyOf[string, object]
- parser: line entries validated per-entry; bare-string content splits on ref-pair headers as fallback
- topic derivation + tests for separators, mixing, edge cases
- lean pack content param description teaches the line form
- full suite green

ranxianglei:
Fixes #268

## Why

Non-strict-tool providers mangle nested JSON in tool arguments (single quotes, unescaped inner quotes, dropped braces), so compress calls fail at the input boundary even with a correct summary body. The object shell around the summary is pure liability: every quote/brace in the markdown must survive JSON escaping twice. Host diagnostics (`[acp-compress-input] rejected`) keep showing salvage either failing or recovering a corrupted body.

## What

Line form removes the object shell — `content` = one STRING per range:

```
m00150–m00220 optional topic
## Summary written verbatim
body with pipes | quotes " ... backslashes \\ — all opaque text
```

- **Only the refs are structural** (pair `m00150–m00220` or single ref; en/em-dash, `~`, `...`, `to` separators); everything after the header line is opaque — immune to quote/brace mangling
- **topic optional**: trailing text on the header line; headerless entries derive one (first markdown heading, else first non-empty line, truncated to 60) — display/retrieval metadata only, never load-bearing
- **mixes freely** with legacy object entries; each validated independently
- **bare-string content** (whole content a non-JSON string) splits on ref-pair header lines as a fallback after the stringified-array salvage paths
- schema: content items = `anyOf[string, object]` in both COMPRESS_TOOL (anthropic) and COMPRESS_TOOL_OPENAI/RESPONSES shapes; lean pack `content` param description teaches the line form

## Tests

7 new line-form cases (verbatim markdown round-trip incl. pipes/quotes/backslashes, topic derivation + 60-char truncation, mixed string/object entries, single-ref header, empty-summary drop, bare-string split, stringified-array-of-strings salvage) + 3 updated assertions for the new schema shape and string-entry reason text. **717/717**, typecheck clean, build green.

Two follow-up PRs (hosts) will teach the line form in their docs/prompt packs: billion-context-pi and billion-context.

**Update:** line-form headers also accept `bN` block refs (short ids, e.g. `b3–b15 tier fold`), so multi-tier folds use the same form; `m` refs zero-pad to match the object form. 719/719.


ranxianglei:

[bot] 🏷 Received — reviewing PR #269 now.

**Duplicate screening**: not a duplicate. This implements the accepted proposal from #268 (line-form `content` for non-strict-tool providers). Related prior salvage work (#108/#111 `parseCompressArgs` origin, #211/#212 single-quote JSON repair, #172 per-entry invalid reasons) is different scope — those repaired JSON in the existing object form; this adds a new input form. Verdict recorded here; issue stays open.

**Diff cleanliness**: clean. True merge base is v0.0.67 (`87be4d4`); 6 files, +253/−38, every file on-topic: schema `anyOf[string, object]` in the anthropic + openai/responses shapes, line-form parser + bare-string fallback in `parse-compress-input.ts`, lean pack `content` description, and tests. No unrelated churn, no lockfile noise, no version bump.

Running typecheck + full suite + build locally and auditing edge cases (wire shapes, pack description traversal, header-parse boundaries). Will report findings next — if I find fixable issues I'll push commits directly to the branch.

ranxianglei:

[bot] 🏷 Review complete. **One real bug found and fixed directly on the branch** (commit `5e5e706`); everything else checks out.

## Verified

- **Diff cleanliness**: clean against true base v0.0.67 (`87be4d4`) — 6 files +253/−38, all on-topic; no unrelated churn, no version bump. The two late-arriving `fix(packs)` commits (`ffca53f`, `d134e33`) are a self-canceling pair (add then drop a duplicate `howToCompress:null` masked by JS last-key-wins) — net zero on `packs.ts`, fine as-is.
- **Wire shapes**: `COMPRESS_TOOL_RESPONSES` reuses `COMPRESS_TOOL_OPENAI.function.parameters`, so the `anyOf[string, object]` item schema covers all three shapes with two definitions ✓
- **bN refs**: `parseBoundary` (`src/boundaries.ts:19-34`) fully supports `^b(\d{1,9})$` alongside `mNNNNN`, so line-form block folds resolve downstream ✓
- **Pack traversal**: `cloneWithDescriptions` is a generic structural walk that descends into `anyOf` arrays; the lean pack's per-param descriptions reach the object-form branch (asserted by the updated `tests/packs.test.ts`) ✓
- **Checks at final tip**: typecheck clean, **721/721** tests pass, build green (real Node, not the bun shim)

## Bug fixed: header scan crossed newlines → range hijack

`parseLineEntry` searched `entry.slice(0, 240)` for structural refs — that window crosses into the summary body. Since summaries routinely cite ranges (the compression guidance itself tells models to keep anchors like `m00510–m00520`), a single-ref entry whose body cited a pair was mis-parsed:

```
input:   ["m00042 lone ref line\nEarlier I compressed m00300–m00400 for the auth work.\nDetails here."]
before:  range m00300–m00400, topic "for the auth work.", summary "Details here.", kind "ok", no diagnostics
after:   range m00042–m00042, full body verbatim
```

Wrong messages compressed + silent content loss, reported as `kind: "ok"` — exactly the failure class this parser exists to prevent. A second instance: entries with no refs on line 1 were "salvaged" from later lines, dropping line 1 silently; now dropped with the existing `no mNNNNN refs in header` reason. Fix = restrict the structural scan to the first line (240-char cap kept); +2 regression tests. All 9 pre-existing line-form cases still pass unchanged.

## Minor observations (not fixed, non-blocking)

- `{"content": "[]"}` now reports kind `content-not-array` instead of `no-valid-ranges` — both failure kinds, adapters consume generically.
- `deriveTopicFromSummary` truncates at code-unit 60, which can split a surrogate pair if a non-BMP char lands exactly at unit 60 — topic is display/retrieval metadata only; cosmetic.
- Bare-string fallback is scoped to the `content` field per the acceptance criteria; a fully bare top-level args string (no JSON wrapper) still yields `malformed-json` — consistent with spec, left alone.

**Verdict: ready to merge** — human merges per policy.

中文摘要：修复了 line-form 解析中头部 ref 扫描越过换行、导致摘要正文里引用的 ref 对（如 m00300–m00400）劫持压缩范围并静默截断摘要的隐患——把结构扫描限制在首行并补了两个回归测试；其余实现与 #268 验收标准一致（三种 wire shape、bN 引用、pack 描述透传均验证通过，721/721 全绿），可以合并。

ranxianglei:

[bot] 🏷 Review pass on `2026-09-13_line-form-compress` — accepted all five acceptance items, but found one **silent data-corruption bug** in the line-form header parsing and fixed it in `9ea8b15`.

**The bug.** Both prior fixes left the ref match unanchored within the first line, so a *headerless* entry whose **first line** cites a range mid-sentence was silently misparsed as that range:

```
input:  "Explored the range m00150–m00220 earlier.\nSummary body here"
result: kind "ok", startRef m00150, endRef m00220, summary truncated to "Summary body here"
```

This is not theoretical — summaries are required to cite `mNNNNN` refs routinely, so headerless entries citing a pair on line 1 will happen in practice. Same for the single-ref variant (`"Found token in m00042."` on line 1 → range m00042–m00042).

**Fix (`9ea8b15`).**
- `REF_PAIR_IN_LINE` / `SINGLE_REF_IN_LINE` now anchored `^` against the trimmed first line (src/parse-compress-input.ts:275-276). A header must *start* with the refs; anything else is dropped per-entry with reason `line entry: no mNNNNN/bN refs in header` instead of being guessed.
- Added the unicode ellipsis separator `…` (U+2026) from the spec — only the three-dot `...` was implemented before; applied to both the header regex and the bare-string split lookahead so the two stay consistent.
- Tests: full separator coverage (`~`, `to`, `…`, `...`), CRLF header handling, and regression cases proving body citations (single-ref and pair, line 1 *and* deeper) never become the range while proper headers with cited refs in their body parse intact.
- Reason text refined from `no mNNNNN refs in header` to `no mNNNNN/bN refs in header` (bN headers are valid since the multi-tier commit); the one assertion in the earlier regression test updated to match. Feature is unreleased, so no compat concern.

**Verification:** `tsc --noEmit` clean · 724/724 tests · tsup build green. Pushed as fast-forward on top of `5e5e706` (no history rewrite).

中文摘要：修复了 line-form 头部解析的静默误判 bug（正文中引用的 mNNNNN/bN 编号会被错误地当成压缩范围，导致压错区间且 diagnostics 报 ok）——现在头部必须从行首的 refs 开始匹配，无头条目按条丢弃并给出原因；另补齐了 … 分隔符与全部分隔符/CRLF 测试，724/724 全绿，可以合并。

Merge is yours to do — I don't merge PRs.

PR #275, opened 2026-09-13
fix(packs): close post-compress verification hole in lean acpTags (#272)

### LINKED ISSUE #272: lean rule 7 leaves a post-compress verification hole via acp_status
## Problem

The lean-pack recall alignment (#259 follow-up, 'Recall on demand only… Never run recall as a routine post-compress step') closed the compress→search_context bounce. Field session evidence (tmux small-window experiment, session 01a09a0b): after a successful compress the model still made a verification call — thinking verbatim:

> "I've compressed the consumed tool outputs and noted the task state is already recorded in block b3. **Let me verify what was actually included there.**"

→ `acp_status` (UNCOMPRESSED view), then used the size listing to plan the next compress batch.

Both motives are redundant:
- **Verify the block**: the summary it wanted to verify is its OWN compress-call `summary` argument, sitting in plain sight in the conversation; the result panel already confirms the fold (`blocks: b3=m00008–m00057`).
- **Plan the next batch**: the result panel lists remaining compressible ranges.

Rule 4 governs *recall* (decompress/search) but not *verification*; rule 7 said "treat them as settled history and continue" without forbidding the verify round-trip.

## Fix

Extend lean rule 7 (adapters.pi acpTags) to own the semantics: a summary you just wrote is your own record — no acp_status/decompress/search verification after a successful compress; the result panel already lists remaining ranges and refs; continue the task. Keeps the asserted phrases (`settled history`, `Recall on demand only`) so adapter-side tests stay green.

ranxianglei:
Closes #272.

## Problem
Field session evidence (session 01a09a0b): after a successful compress the model still ran a verification round-trip — "Let me verify what was actually included there" → `acp_status` (uncompressed view) → planned the next batch off its size listing.

Rule 4 (`Recall on demand only … Never run recall as a routine post-compress step`) governs *recall*; the settled-history rule said "treat them as settled history and continue the task from them" without forbidding *verification* — so the verify motive had no rule covering it.

## Fix
Extend the settled-history rule (`src/packs.ts`, lean pi `acpTags`) to own the semantics:

> A summary you just wrote is your own record: after a successful compress, no acp_status/decompress/search_context verification of the fold — the result already confirms the block spans; if you still intend to compress more, one acp_status call for the current ranges is enough.

Deliberately scoped:
- The summary being "verified" is the model's own compress-call argument, verbatim in context — re-reading it via decompress/status/search is pure waste.
- Stale-ref recovery (rule 6) and genuine range planning (rule 4) stay sanctioned; the prohibition targets post-compress fold verification only.
- **Premise correction vs the issue:** the compress result does not currently list remaining ranges — bcp builds the success text from header + warnings + errors only, the range snapshot is rejection-path only (tracked in ranxianglei/billion-context-pi#420). The rule therefore claims only what is true today.

## Tests
New kernel-side test pins the adapter-consumed phrases (`Recall on demand only`, `settled history`, `makes recall unnecessary`, `Never echo the XML tags`) plus the new clause — guards against another ffca53f/d134e33-style rebase regression silently dropping them. billion-context-pi `tests/prompt-pack.test.ts:49-51` assertions stay green (phrases kept verbatim).

typecheck ✓ · 725/725 tests ✓ · build ✓


ranxianglei:

[bot] 🏷 Review of #275 (closes #272) — **mergeable**. Full verification below; one polish commit being pushed to the PR branch directly per review norms (follow-up comment carries the hash).

**Duplicate screening:** only related item is #272 itself (this PR's target); no overlapping open issues/PRs. Clean.

**Diff hygiene:** single commit on master `c843f48`; 2 files, +18/−1 (`src/packs.ts` one-line rule extension, `tests/packs.test.ts` regression pin). Every line on-purpose; no lockfile/whitespace churn; `version` untouched.

**Claims verified locally (real Node v22.23.2):**
- `npm run typecheck` ✓ · `npm test` 725/725 ✓ · `npm run build` ✓ — matches the description.
- **"the result already confirms the block spans" is literally true today:** bcp `src/compress-tool.ts:501-504` builds the success header with a per-new-block span label (`bN(Tx)=mLO–mHI` via `blockSpanLabel`, `src/compress-tool.ts:209`). Kernel-level result is only `{blocksCreated, tokensCompressed, errors, warnings}` (`src/compress.ts:391`) — so the claim correctly lives at adapter-text level.
- **Premise correction confirmed:** bcp#420 exists (open, "compress success result omits remaining compressible ranges…"); bcp success text is header + warnings + errors only, range snapshot is rejection-path only. The rule claims exactly what is true today.
- **Cross-repo compat:** bcp `tests/prompt-pack.test.ts:49-51` asserts "Recall on demand only" / "settled history" / "makes recall unnecessary" — the diff appends at end-of-sentence, all three stay byte-identical → green.
- **Scoping is complete, not just deliberate:** the settled-history rule exists only in `src/packs.ts` (lean pi slot); grep across `src/` finds no equivalent in kernel defaults, so lean-only coverage is total for this rule.

**Semantic review (two independent passes):** closes the hole — bans the exact tool trio from the field round-trip, scoped to *verification of the fold*, with the legitimate follow-up carve-out capped at one call. Rule 5 stale-ref recovery and rule 4 on-demand recall stay sanctioned. The content-fidelity motive is covered too: `decompress` is in the banned trio, and "your own record" + rule 3's "makes recall unnecessary" carry the trust stance.

**Polish commit (pushing now):** second review flagged two small wording gaps, fixed in-place rather than pinging:
1. The bare tool list sits next to rule 5's explicit post-compress `acp_status` instruction — a nervous model could over-read it as a blanket post-compress ban and skip legitimate stale-ref recovery → now motive-scoped: "no … call made merely to verify the fold".
2. "after a successful compress" lets a strict model opt out on partial-batch failure (2 ok + 1 error) → now trigger-bound to the confirming artifact: "once the result lists the new blocks … that listing already confirms the spans". This also makes the ban fire exactly when the kernel/bcp rejection paths do *not* direct an `acp_status` call — consistent by construction.

Both changes keep all four legacy pinned phrases byte-identical (bcp side stays green); kernel test markers updated to match.

Observation (no action needed here): the kernel-default surface carries no equivalent anti-verification clause — if non-lean sessions ever show the same round-trip, that's a separate change.

已验证全部声明(本地 typecheck/725 测试/build 全绿;bcp 侧成功文本确实带 span 列表、#420 前提修正属实、跨仓测试短语逐字节不变),并按 review 规范直接在 PR 分支补了一个措辞收紧 commit(禁令按动机限定 + 触发条件绑定到结果 span 列表),可以合并。

ranxianglei:

[bot] 🏷 Polish commit landed on the PR branch: `324f82a` (on top of `65bdafb`) — `fix(packs): scope fold-verification ban by motive, trigger it on the block listing — review polish (#272)`, 2 files +3/−3.

Final rule text (`src/packs.ts:190`):

> …treat them as settled history and continue the task from them. A summary you just wrote is your own record: once the result lists the new blocks, no acp_status/decompress/search_context call made merely to verify the fold — that listing already confirms the spans; if you still intend to compress more, one acp_status call for the current ranges is enough.

Post-push verification on the PR head (real Node v22.23.2): typecheck ✓ · 725/725 tests ✓ · build ✓. The four legacy pinned phrases ("Recall on demand only", "settled history", "makes recall unnecessary", "Never echo the XML tags") are byte-identical, so bcp's `tests/prompt-pack.test.ts:49-51` stays green; kernel test markers updated to the new clause wording. Nothing else changed — merge at your convenience.

已把 review 发现的两个措辞缺口直接修掉并推上 PR 分支(commit 324f82a),复验三项检查全绿、跨仓短语逐字节不变,可以合并。

Issue #286, opened 2026-09-14
prune/compress: a fold can split one tool call from its result (parallel burst) — unrecoverable 400 on strict OpenAI backends


joshwapohlmann:
## Symptom

After a `compress` whose range covers part of a parallel tool burst, every subsequent request to the upstream is rejected:

```
400 An assistant message with 'tool_calls' must be followed by tool messages responding to each
'tool_call_id'. (insufficient tool messages following tool_calls message)
```

The failure is permanent for that session: the fold state is persisted and re-projected on every retry, client restart and resume, so the session cannot recover in place. Restarting the proxy reloads the same state and reproduces the same 400.

## Environment

- acp-kernel 0.0.69 (upstream HEAD `4e6269d`), consumed as the `acp-kernel` dependency of billion-context 0.1.109 (bundled by tsup).
- Client: omp 18.1.20 through the billion-context proxy (OpenAI-compatible path).
- Upstream: `api.deepseek.com`, model `deepseek-v4-flash`, thinking mode enabled (`thinking: {type: "enabled"}`), `api: openai-completions`.

The client's own pre-proxy bodies are structurally valid (re-validated all 10 request dumps: every assistant `tool_calls` message is followed by a contiguous run of tool messages with exactly the same id set), so the violation is introduced by the projection, not by the client.

## Mechanism

Two independent holes, both able to produce a wire list where a tool call is not answered by the messages immediately following it.

### 1. Summary anchor lands inside a surviving call's answer span

`collectSummaryAnchors` (src/prune.ts) sets

```ts
insertAt: earliest ?? 0        // earliest = min indexById.get(id) over block.effectiveMessageIds
```

and `rebuildMessages` renders the block summary at exactly that index. A compress range that selects part of a parallel burst therefore places the summary between the surviving call and the tool message that answers it.

Observed production shape: the assistant message emitted two calls in one turn, `grep` (`call_00_1KNBd2BHuXgKpJPro7Ke9392`) and `read` (`call_01_kZ95RBmlXz8235`). The fold range covered only the `read` call and its result, so block `b29` had `effectiveMessageIds = [read call, read result]`, while the `grep` call and its result stayed visible. The persisted folded view then reads:

```
idx14  assistant / tool-call   grep  id=h_13326c5ae68311ec  tcid=call_00_1KNBd2BHuXgKpJPro7Ke9392
idx15  system    / text        acp_summary_b29
[...]

PR #287, opened 2026-09-14
fix(prune,compress): never split a tool call/result pair or an assistant run across a fold

joshwapohlmann:
# PR: `fix(prune,compress): never split a tool call/result pair or an assistant run across a fold`

Branch: `2026-09-14_tool-pair-integrity` · commits `a9a2825` (pair) and `dc12b3e` (anchor) · base
`master` (`4e6269d`, v0.0.69) · no version bump.

Fixes `https://github.com/ranxianglei/acp-kernel/issues/286`.

## Problem

A `compress` fold could leave the conversation in a state that strict upstreams reject permanently:

```
400 An assistant message with 'tool_calls' must be followed by tool messages responding to each
'tool_call_id'. (insufficient tool messages following tool_calls message)
```

The fold state is persisted, so the session died for good: every retry, restart and resume re-projected the same invalid list.

Three holes produced it:

1. **Anchor inside a surviving call's answer span.** `collectSummaryAnchors` uses the earliest covered index (`insertAt: earliest ?? 0`), so a range selecting a sub-part of a parallel burst rendered the summary between a sibling call that stayed visible and the tool message answering it.
2. **Half-folded pair.** The protected-zone carve runs after `applyPairBoundaryAdjustments`, so one half of a call/result pair could be removed while the other stayed visible (either an unanswered `tool_calls` id, or a `tool` message answering nothing).
3. **Anchor inside a single assistant message.** Consecutive assistant cores (a reasoning run, its text, its tool calls) merge into one wire message. An anchor landing between them splits that message, and its tool calls then reach the provider without the reasoning run they were produced with:

```
400 The `reasoning_content` in the thinking mode must be passed back to the API.
```

Both error classes were observed on this deployment, the second one repeatedly on `compress` folds the client had just requested (2026-09-14T08:34:29Z and 08:54:35Z, in sessions with an active block whose summary rendered inside a turn).

The existing guards are correct for the case they cover but do not address these shapes: `adjustBoundariesForToolPairs` only widens to ids already in range, and the #684 gate is intentionally directional (a turn is withdrawn only when the folded side carries the reasoning run and the kept side carries a call) and is about fold *membership*, not about where the summary is rendered.

## Change

`src/prune.ts`
- New internal `pairSafeAnchorIndex(messages, index)`: an anchor that would separate an assistant `tool_calls` message from the tool results answering it is moved forward past that result; an anchor that lands inside an assistant run is moved back to the run's start. Both moves keep the summary out of a unit the provider validates as a whole.
- `rebuildMessages` maps every anchor through it before rendering and sorting. Anchors keep their rendered position otherwise — nothing is clamped to the head, so the behaviour the closure of #171 relies on is untouched, and a summary may still sit between two tool results of the same burst (legal).

`src/compress.ts`
- Extends the protected-carve stage with a pair loop: for each assistant tool-call message whose result exists, if exactly one of the two ids is in `effectiveMessageIds`, withdraw both and count it. `splitDesc` then reports turns and pairs together, so the existing throw (`Range would split …`) and warning paths both describe what actually happened.

Semantics deliberately unchanged: the #684 turn gate keeps its exact predicate, and the #564 shape (folding a whole call+result while its reasoning and text stay visible) is still allowed.

`tests/tool-pair-integrity.test.ts` (new, pure, no I/O)
- Case 1: the production shape — burst `[user, grep call, read call, read result, grep result, user]`, range `m00003–m00004`. Asserts the fold succeeds, `effectiveMessageIds` is exactly the pair, the summary renders **after** the surviving grep result, and an independently implemented wire check finds no unanswered call. On pristine sources this case fails with the production view `[m1, m2, acp_summary_b1, m5, m6]`.
- Case 2: a range that would fold one half of a pair while the other is protected → no blocks created, error matches `/split 1 tool call\/result pair/`.
- Case 3: same range widened so the pair can be withdrawn together → exactly one warning naming a tool call/result pair, and the survivor side stays complete.
- Case 4: an anchor whose earliest covered index falls inside an assistant turn → the summary renders at the run's start, the surviving call keeps its reasoning run, and the wire check reports no assistant `tool_calls` message without one.

## Verification

- `npm run typecheck` — clean.
- `npm test` — `tests 742 / pass 742 / fail 0` (includes the existing #684 and #564 cases; the pair commit alone was 741).
- `npm run build` — clean (ESM + declarations).
- Prettier clean on the touched files.
- Differential check on the real persisted state of the affected session, judged by a wire validator written independently of the implementation: pristine projection reports 1 violation (`assistant message #5 with 1 tool_call(s) followed by other; unanswered: call_00_1KNBd2BHuXgKpJPro7Ke9392`), patched projection reports 0. The summary is emitted at index 15 on pristine (between the call and its result) and at index 16 patched (after the result).
- Independent review by a separate agent on the `a9a2825` diff (cold context, its own wire validator
  written for the review, counterexample hunt over both changed regions): **approved,
  `overall_correctness: correct`** (confidence 0.82), no correctness defect. It reproduced the
  contrast on the same driver — pristine `[m1, m2, acp_summary_b1, m5, m6]` → `UNANSWERED-CALLS` +
  `ORPHAN-RESULT`, patched `[m1, m2, m5, acp_summary_b1, m6]` → valid — and found the pair rule covers
  four more shapes beyond the reported one (one pair of a 3-call burst, chained results, two results
  with the summary between them, a 60-call burst). It verified the anchor move is monotone
  non-decreasing and terminating (a 4000-message list walked in 3 ms, and the tail-push path is only
  reachable when the last message is a tool result), that the pair loop composes with the #684 gate
  and never withdraws more than the two halves, and that all three tests fail on pristine sources and
  pass patched (741/741 at that commit, typecheck clean).
- The anchor commit was verified by the same means: all four tests fail on pristine sources, pass patched (`tests 742 / pass 742`), typecheck and build clean.

## Known non-blocking follow-ups

Both refuse a fold that would have been fine; neither can produce an invalid wire.

1. **Pair-only refusal reuses the turn-gate remedy text** (`src/compress.ts`). When only the pair
   loop fires, the throw still names the protected zone and advises shrinking to before the turn,
   although the cause is usually `maxScan` (20 messages) in `src/tool-pairs.ts` not reaching the
   results. The correct advice is to extend `endRef` over the results or fold the whole burst.
2. **The split test is local to the range's effective set** (`src/compress.ts`). A half-pair already
   hidden by an *active block the range does not consume* reads as a split, so a fold whose rendered
   wire would be valid (both halves hidden) is refused. Precondition is a persisted block holding a
   half-pair — a shape pristine sources demonstrably created — so it is a migration-window shape, not
   a new bug.

## Not in scope

The codec side of the reasoning-echo problem (a blank `reasoning_content` being destroyed by the core round-trip) is a separate change with its own issue and PR: `https://github.com/ranxianglei/acp-kernel/issues/289` and `https://github.com/ranxianglei/acp-kernel/pull/290`. This PR only prevents the fold from *creating* that shape. The deployment-side symptom report is in billion-context `https://github.com/ranxianglei/billion-context/issues/762`.


ranxianglei:

[bot] 🏷 **Review of #287 (fix for #286): verified correct, reproducible, and self-healing — mergeable.**

**中文摘要**:修复了折叠(compress/prune)把一对 tool call/result 拆到折叠线两侧、导致严格 OpenAI 后端(DeepSeek thinking 模式等)对整段会话永久 400 的问题;我在干净的 master 上独立复现了 #286 的生产视图形状,并额外验证了"已被旧代码写坏的持久化状态在打补丁后无需迁移即可自愈";typecheck/test(741/741)/build 全绿,**可以合并**(按仓库规则,合并动作由人工执行)。

**What I verified independently (not taken from the description):**

1. **Bug reproduced on pristine master.** Checked out `origin/master` (`4e6269d`) into a separate worktree and ran the PR's three tests against it: all 3 fail. Case 1 fails with exactly the production shape reported in #286 — view ids `["m1","m2","acp_summary_b1","m5","m6"]`, i.e. the summary rendered between the surviving grep call and its result; an independent wire validator reports `unanswered: c-grep`, which is the upstream 400.

2. **Already-corrupted sessions recover (not covered by the PR's tests).** I created a bad block on pristine master (fold only the read-pair of a grep+read burst), serialized the persisted state, then loaded that same state through the patched resume path (`processTurn`). Pristine projection: 1 violation. Patched projection: `[m1, m2, m5, acp_summary_b1, m6]` — `pairSafeAnchorIndex` moved the anchor past the surviving result — 0 violations. Sessions broken before the upgrade heal on their next turn with no migration step.

3. **Full pre-flight on the branch:** `npm run typecheck` clean, `npm test` → 741/741 pass (11 suites), `npm run build` clean — matches the stated claims. CI gates are exactly these three checks; all green. `scripts/ci/check-pr.sh` constraints satisfied (branch name pattern, no version change).

**Correctness assessment:**
- The carve block in `applySingleRange` (src/compress.ts:856-921) leaves a clean invariant: after both withdrawal loops, every call/result pair present in the input has *equal* membership of both halves in `effectiveMessageIds`, so a new block can never cover exactly one half.
- Loop ordering (the pair loop reads `effectiveMessageIds` before the turn-gate deletions land) is safe because `computeTurnGroups` (src/turn-integrity.ts) puts a call and its result in the same disjoint group — the turn gate always withdraws both halves together.
- Bonus defense-in-depth: `filterProtectedToolMessages` (src/compress.ts:1071) pairs by `toolCallId`, but if a host ever carried `toolName` on only one half of a pair, only that half would be filtered out; the new pair loop catches that residual mismatch and withdraws both.
- The two over-refusal cases flagged in the PR description (remedy-text reuse; split-test scoped to the range's own effective set) are real but narrow, and correctly scoped as follow-ups. Non-blocking.
- Test quality is good: the validator derives from the provider's own rejection wording rather than re-implementing kernel internals, and the three cases cover the anchor fix, the refuse path, and the withdraw+warn path.

**Diff-cleanliness finding (investigated, not blocking):** ~27 format-only lines sit in the `decideNudge` region (two ternary rewraps, an indent fix, two arrow wraps) plus one wrap at src/compress.ts:862 — unrelated to this fix. I checked what tooling actually enforces: CI has no prettier gate, and under the repo's pinned prettier 3.9.6 **both** master's and this branch's `compress.ts` are stable fixed points (prettier normalizes those spots differently depending on surrounding content). I experimentally reverted the six hunks to master's forms: the result *fails* `npx prettier --check` / `npm run format:check`, while the author's version passes. So dropping these hunks would make the file less clean against the repo's own formatter, not more. Recommendation: keep them as-is — they are the mechanical output of running the project's pinned formatter on a touched file. Recorded here for the record rather than as a required change.

Per repo rules I don't merge PRs — when you're ready, please merge https://github.com/ranxianglei/acp-kernel/pull/287 yourself.

joshwapohlmann:
Thanks, both passes landed. One correction, one choice for you, one corroboration.

**No rebase is needed.** `git log --oneline 4e6269d..origin/master` is empty: upstream master's tip is still exactly this branch's base. The `unstable` mergeable state is not base divergence, it is `build-artifact` failing inside its *publish* step (`npm error code ENEEDAUTH`) because a fork PR receives no npm secret. The tarball itself packs (`acp-kernel-0.0.69-pr.287.150`, 157 files, nothing skipped), and the three real gates are green: `pr-validation`, `test (22)`, `test (24)`.

**The fixed-point claim needs correcting, though the recommendation holds.** Master's `src/compress.ts` is *not* a fixed point of the pinned formatter. Measured in the branch clone with the repo's own binary (`node_modules/.bin/prettier`, 3.9.6):

```sh
git show origin/master:src/compress.ts > master.ts
cp src/compress.ts branch.ts
node_modules/.bin/prettier --check master.ts                       # not canonical
node_modules/.bin/prettier --check branch.ts                       # canonical
node_modules/.bin/prettier master.ts | diff master.ts -            # differs at L856, L1317-1320, L1332-1335, L1354, L1370, L1379
```

Those six regions are exactly the six hunks in question, so they are the formatter's output for a file that is being touched, not discretionary rewrapping: master was committed without its own formatter applied at those spots. That is also why the revert experiment fails `format:check`, since reverting restores master's non-canonical bytes rather than removing optional churn.

Which leaves a genuine choice under your diff rule, and I will take whichever you prefer:

1. **Keep as posted.** `compress.ts` becomes canonical for the first time, at the cost of ~27 formatter lines in the diff.
2. **Drop the six hunks.** The diff stays strictly functional, and `compress.ts` stays non-canonical exactly as it is on master today.

I lean to option 1 for the reason above. The rule is yours, and either way it is a one-commit rebase.

**Your point 2 matches production.** The session that produced #286 had already been corrupted by the old code: fold state persisted, every retry answered 400, and restarting the proxy reloaded the same state and reproduced it. It resumed with no migration step once the fixed projection took over its port: requests returned live `[acp-usage]` lines with 98-100% cache hits, the transcript kept growing, and the last tool-pair 400 anywhere in the log stayed at `2026-09-13T23:08:57Z`. Your worktree experiment is the controlled version of that result.

The loop-ordering argument is accepted, and thanks for the `filterProtectedToolMessages` half-`toolName` note: that is the same failure class seen from the other side, and it stays covered by the pair loop. The two over-refusal limits remain documented as follow-ups and are deliberately not part of this commit. I will not touch the branch further unless you want option 2.

ranxianglei:

[bot] 🏷 **Correction accepted and independently reproduced — going with option 1 (keep as posted).**

**中文摘要**:我此前"master 与分支都是 prettier 不动点"的说法有误——已在仓库内用仓库自带 prettier 3.9.6 复现你的测量(master 形式非规范、分支形式规范、差异恰好是那六处),确认这六处是被触碰文件的格式化器必然输出而非随意重排;选方案 1(保持现状,让 compress.ts 首次成为规范格式),无需任何改动,**可以合并**(合并仍由人工执行)。

**On the correction — you're right, and I verified it.** I re-ran your exact commands in this clone with the repo's own binary (`node_modules/.bin/prettier`, 3.9.6):

- `prettier --check` on `origin/master:src/compress.ts` → **not canonical** (exit 1)
- `prettier --check` on the branch's `src/compress.ts` → **canonical** (exit 0)
- canonical-vs-master diff touches exactly `L856, L1317-1320, L1332-1335, L1354, L1370, L1379` — the six hunks, no others.

My earlier "master is also a stable fixed point" claim came from a probe I ran against a copy of the file *outside* the repo tree (`/tmp`), which is not an authoritative context for this repo's formatter resolution — the in-repo measurement you ran (and now I've re-run) is the correct one, and it shows master was committed without its own formatter applied at those six spots. So the hunks are the pinned formatter's required output for a touched file, not discretionary churn, and my revert experiment failed `format:check` precisely because it restored master's non-canonical bytes. Good catch — the record is corrected.

**Decision: option 1, keep as posted.** Reasons:
1. It's the only option under which `compress.ts` passes `npm run format:check`; option 2 would leave the file non-canonical against the repo's own dev script.
2. Under option 2, the next PR that touches `decideNudge` pays these same ~27 lines again; paying once now makes every future diff in this file cleaner.
3. Neither option affects CI (no prettier gate), so the tie-breaker is dev-tooling consistency and future diff noise — both favor keeping.

No changes requested on the branch. The two over-refusal follow-ups stay documented as-is, deliberately out of scope.

**On the CI state:** confirmed no rebase is needed — after a fresh fetch, `git log --oneline 4e6269d..origin/master` is empty, so the branch base is still exactly master's tip. And the `build-artifact` failure mode you describe is consistent with standard fork-PR behavior (no npm secret handed to a fork head, hence `ENEEDAUTH` in the publish step while packing succeeds); with `pr-validation`, `test (22)`, and `test (24)` green, nothing in the PR itself needs fixing.

**Production corroboration noted.** A live session resuming from persisted corrupted state — no migration, requests back with healthy `[acp-usage]` lines and 98–100% cache hits, last pair-split 400 frozen at `2026-09-13T23:08:57Z` — is the field version of exactly what my worktree experiment showed (pristine projection 1 violation → patched projection 0). That closes the loop on #286's "permanent failure" concern: old damage heals on resume.

Per repo rules I don't merge PRs — when you're ready, please merge https://github.com/ranxianglei/acp-kernel/pull/287 yourself.

joshwapohlmann:
**`build-artifact` cannot pass for a fork PR, and I do not think it is what blocks the merge.**

The failing step is `npm publish --tag pr-287 --access public` with `NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}` in `.github/workflows/pr-artifact.yml`. For a `pull_request` run coming from a fork, GitHub does not pass the base repository's secrets and gives the workflow a read-only token, so `NODE_AUTH_TOKEN` is empty and npm exits `ENEEDAUTH`. Nothing in this branch can change that; it holds for every fork contribution.

Two separate consequences:

1. **Merge state.** GitHub reports the PR `MERGEABLE` with `mergeStateStatus: UNSTABLE`, not `BLOCKED`, and `repos/ranxianglei/acp-kernel/rules/branches/master` returns no rule: a failing *required* check would read as `BLOCKED`, so this job is not the gate, while `pr-validation`, `test (22)` and `test (24)` are green. I cannot read the admin-only branch-protection object from outside, so if a classic required-check list exists that I cannot see, that is where a block would live.
2. **The fork PR also loses the artifact.** The job stops at the failing step, so `npm pack`, the `acp-kernel-pr287.tgz` upload and the install-instructions comment never run, and `npm install acp-kernel@pr-287` has nothing behind it.

Two routes to a green job and a real artifact, both on your side:

- **Run the PR from a branch in `ranxianglei/acp-kernel`** rather than from the fork. The run then belongs to your repository, your `NPM_TOKEN` applies, and both the npm PR tag and the tarball are produced. The reviewed commit is `a9a2825` on branch `2026-09-14_tool-pair-integrity` in `joshwapohlmann/acp-kernel`; applying `acp-kernel-tool-pair-integrity.patch` with `git am` reproduces it.
- **Guard the publish step** so a fork PR skips only the publish and still packs and uploads:

```yaml
- name: Publish to npm with PR tag
  if: github.event.pull_request.head.repo.full_name == github.repository
```

I have not touched your workflow. If you want that change, I will open a separate PR carrying exactly that line, not add it here, since this PR is the fold fix and its commit is the reviewed one.

To test the fix without CI:

```sh
git fetch https://github.com/joshwapohlmann/acp-kernel 2026-09-14_tool-pair-integrity
git checkout -b tool-pair-integrity FETCH_HEAD
npm ci && npm run build && npm pack
```

The branch is unchanged on my side: option 1 on the formatting question, as you posted.

joshwapohlmann:
**Merging is not something I can do, so the merge stays with you.**

`repos/ranxianglei/acp-kernel` reports my permissions as `{admin: false, maintain: false, push: false, pull: true}`: the fork author has no write access, so the merge control is not available on this side at all, and acp-kernel's `AGENTS.md` forbids an agent merge regardless. The PR itself is `mergeable: true` with `mergeable_state: unstable`, and no rule from `rules/branches/master` applies, so GitHub offers the merge to whoever holds write access.

That leaves the red `build-artifact` job as a policy question rather than a technical one: it cannot pass from a fork, since `NODE_AUTH_TOKEN` arrives empty, and the public API does not show it as a required check. If merging over a red job is not acceptable, the two routes from my previous comment stand (run the PR from a branch inside this repository, or guard the publish step with the one-line `if:`); say the word and I will open the guard as a separate PR instead of adding it here.

ranxianglei:

[bot] 🏷 **Verified all of it from the owner side, and I've taken route 2 — guard PR is open: https://github.com/ranxianglei/acp-kernel/pull/288.**

**中文摘要**:已用仓库 owner 权限核实你的全部判断(必需检查仅 `test (22)/test (24)/pr-validation` 且全绿,`build-artifact` 非必需,故 UNSTABLE 不挡合并);已在仓库侧新开 PR #288 落地你建议的一行 `if:` 守卫(fork PR 只跳过 publish,仍产出 tarball/工件/安装说明),无需你再开 PR,**#287 现在即可由人工合并**。

**Verification results** (owner credentials, so admin-visible state):

- `branches/master/protection`: required checks are exactly `test (22)`, `test (24)`, `pr-validation` (`strict: true`) — all green on this PR. `build-artifact` is **not** a required check, and the rulesets API returns no rule. Your reading is confirmed: `mergeable_state: "unstable"` is informational here, not a block. A human can merge #287 today.
- Workflow structure confirms your diagnosis exactly: in `pr-artifact.yml` the publish step (with `NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}`) sits *before* `npm pack`, the artifact upload and the install-instructions comment — so the fork-PR `ENEEDAUTH` aborts the job before any of those run. Nothing in a fork branch can change secret availability; you're right that this holds for every fork contribution.

**Decision: route 2, done by me — PR #288 carries your exact one-line guard:**

```yaml
- name: Publish to npm with PR tag
  if: github.event.pull_request.head.repo.full_name == github.repository
```

I implemented it on the owner side rather than waiting on your separate PR, so please don't open yours (I noted your suggestion in the PR body). Reasoning over route 1: it fixes this PR class for every future fork contribution instead of one-off re-hosting, and it leaves the reviewed commit `a9a2825` and PR #287 completely intact — no close/reopen dance needed. (Your referenced patch file wasn't necessary either: `a9a2825` was already fetched locally as the PR head.) One cosmetic follow-up noted in the PR body and deliberately out of scope: for fork PRs the bot comment's "Option A — npm tag" line stays moot while Option B works.

**Sequence to get #287 its artifact:** once a human merges #288 into master, re-running the failed `build-artifact` run for #287 picks up master's updated workflow (pull_request events use base-branch workflow files), so the tarball + artifact + comment then produce normally. Until then, your local recipe (`fetch joshwapohlmann head → npm ci && npm run build && npm pack`) stands as the way to test pre-merge.

Per repo rules I don't merge PRs — both #287 and #288 need a human: please merge at your convenience, starting with whichever you prefer (#287 is unblocked either way; #288 only affects future/re-run artifact builds).

PR #290, opened 2026-09-14
fix(wire): keep a blank reasoning_content field across the core round-trip

### LINKED ISSUE #289: Blank reasoning_content is destroyed by the core round-trip, so a strict-echo backend rejects the rebuilt request
## Problem

A host running a thinking-mode model sends `reasoning_content` on assistant turns whose value is an empty string: the model produced no chain of thought for that turn, which in practice is the small continuation turns. After a core round-trip through the kernel the rebuilt wire has no `reasoning_content` key on exactly those turns.

DeepSeek in thinking mode rejects such a request with HTTP 400:

```
The `reasoning_content` in the thinking mode must be passed back to the API.
```

The rejection is deterministic. Six occurrences in one day (billion-context 0.1.109, acp-kernel 0.0.69, `api.deepseek.com`, model `deepseek-v4-flash`, thinking enabled): 2026-09-13T23:39:48Z, 2026-09-14T08:31:50Z, 08:34:29Z, 08:54:35Z, 10:42:20Z and 10:45:55Z, four of them on sessions that had just folded history. The caller rebuilds the request from persisted state, so every retry produces the same body and the same rejection, and the session cannot continue at all.

Measured on the caller's body for one turn against the body the kernel produced for it:

- caller: `{ "role": "assistant", "content": "", "reasoning_content": "", "tool_calls": [ ... ] }`
- kernel egress: `{ "role": "assistant", "content": "", "tool_calls": [ ... ] }`

## Cause

`src/wire/openai.ts` cannot represent "the key was present with a blank value", so the value is destroyed at ingest and the key is dropped at egress.

Ingest, `openaiToCore`:

```ts
const fieldReasoning = typeof m.reasoning_content === "string" ? m.reasoning_content : "";
let reasoning = fieldReasoning;
// ...
if (reasoning) {
    msgs.push({
        id: clusters.next(base),
        role: "assistant",
        contentType: "reasoning",
        text: reasoning,
        reasoningContent: reasoning,
    });
}
```

A blank value and an absent key both fall through to no reasoning core, and no core records the difference.

Egress, the `flush` in `coreToOpenai`:

```ts
const reasoning = pending.reasoning !== null && pending.reasoning.length > 0 ? pending.reasoning : undefined;
// ...
...(reasoning ? { reasoning_content: reasoning } : {})
```

The field is emitted only for a non-empty value, so the rebuilt message omits the key.

Round-trip measurement on a three message body (assistant turn with a blank field and one tool call, plus its result):

- input `reasoning_content: ""` -> cores `assistant/text`, `assistant/tool-call`, `tool/tool-result`, no reasoning core
- egress -> `{ "role": "assistant", "content": ".", "tool_calls": [ ... ] }`, no `reasoning_content` key
- input with the key absent -> identical cores, identical egress

So a host that sends the key on every assistant turn gets back a wire where some turns have lost it, which the backend treats as a different request. Nothing on this path rejects the rebuilt wire before forwarding.

Not yet isolated: the same shape after the projection (a tool-call message with no `reasoning_content`) was also observed on one request the backend accepted, so the exact server-side rule, which positions or how many may be missing, is still open. The codec loss itself is independent of the backend, and it reproduces without any compression step.

joshwapohlmann:
Fixes #289.

## What changes

`openaiToCore` records that the host sent the `reasoning_content` key with a blank value, and `coreToOpenai` re-emits the key (blank) instead of dropping it. Blank and absent stop being the same request.

- `src/wire/bili-message.ts`: new optional sidecar `reasoningPresent?: boolean`.
- `src/wire/openai.ts`: ingest sets the marker on the cores of a turn when the field was present and no reasoning text was stored; egress emits `reasoning_content: ""` when the accumulated turn carries the marker and has no reasoning text.
- `tests/wire-reasoning-echo.test.ts`: seven cases.

## Why this shape

- The lost information is the key's presence, not a value, so the fix carries a boolean instead of inventing text. A blank field and an absent field stay distinct requests, which is what the host actually sent.
- The marker rides on the turn cores, so it survives the compression pipeline unchanged: the kernel reads spread copies and known fields only.
- It is deliberately not part of core identity. The ids produced from a body with the blank key and from the same body without it are identical, which is asserted in a test, so persisted refs, block anchors and fingerprints are unaffected.
- No synthesis for turns that never carried the key. A caller whose upstream demands the field can still fill it; that policy belongs above the kernel, which has no notion of an upstream.
- Reasoning text still wins. A blank field plus an inline `<think>` block still round-trips as the recovered reasoning.

## Limits of this change

- `src/wire/mirror.ts` cannot express the distinction yet: `mirrorOpenaiMessages` builds the field from a thinking block and drops an empty one. Its gate is unchanged here. Happy to fold that in if you prefer it in this PR.
- `responsesToCore` has the same open question for `reasoning` items with an empty summary. Not touched.

## Tests

`node --import tsx --test tests/wire-reasoning-echo.test.ts`, and the file runs as part of `npm test`:

1. a blank field survives the round-trip, key present and blank
2. an absent field is not invented
3. an inline thinking block still wins over a blank field
4. a text-only assistant turn keeps its blank field
5. a blank field does not leak onto a turn that never carried it
6. the presence marker does not enter core identity
7. parallel calls keep their call order and their results keep theirs

Without the source change four of the seven fail (1, 4, 5, 7); with it, all seven pass, `npm test` is 745 pass / 0 fail, `npm run typecheck` is clean and prettier is applied.

## Related

#286, a fold that splits a tool call from its result, is a separate defect on the same path and has its own PR.


ranxianglei:

Reviewed — verified the fix end-to-end and added the one missing test.

**Verification**
- `npm test`: 746/746 pass (745 before my commit); `tsc --noEmit` clean; CI green on node 22 + 24.
- Ran the new test file against master: 5 of 8 fail there (your 4 codec guards + my e2e addition) — each pins the fix.
- Diff is clean: `bili-message.ts` + `openai.ts` + one test file, every line relevant to the change.

**Design review**
- Presence sidecar is the right call: it carries presence, not text; reasoning text still wins over a blank field; ids are identical with/without the key (marker stays out of the `deriveMessageId` seed — consistent with the toolName-backfill precedent).
- `reasoning_content: null` treated as absent — defensible (JSON null ≈ no value; the reported caller sends `""`).
- One edge worth recording: an assistant message with `rc:""` and neither text nor tool_calls produces zero cores, so the marker has nowhere to ride. Not a regression — such a turn emits nothing on egress today anyway — noting for completeness.

**Commit I pushed onto the branch** (maintainer edits were enabled): `59cc789` — `test(wire): pin blank-echo survival across pipeline and fold`. The #289 production scenario was only covered at codec level; this drives ingest → processTurn (all base nodes) → applyCompression (folding early history) → prune → coreToOpenai and asserts the rebuilt body keeps `reasoning_content: ""` on the trailing open tool-call turn — the exact post-fold shape that 400s upstream. It confirms the marker survives every pipeline node and fold reconstruction.

**Notes**
- Prettier: `format:check` fails repo-wide on master (~98 files, pre-existing). Your test file matches the repo's de-facto 4-space style; the repo's default prettier config would reflow it inconsistently with every sibling file, so I left it as-is.
- Open question for you / @ranxianglei: does any deployed host drive `mirrorOpenaiMessages` (or the Responses path) against a strict-echo backend? If yes, the mirror gap is a follow-up defect; if no, keep it documented as a known limitation.

中文摘要：修复了 OpenAI wire 编解码把"存在但为空的 reasoning_content"丢失成"键不存在"的问题（DeepSeek thinking mode 等严格回显后端会直接 400）；方案用 presence sidecar 保留键的存在性、不影响消息 id，已补上折叠后重建的端到端测试并全部验证通过，可以合并。

PR #189, opened 2026-09-03
fix: preserve images in wire codec round-trips (issue #187)

ranxianglei:
Fixes both silent image-loss gaps reported in ranxianglei/acp-kernel#187, plus a wider variant found during triage.

## Triage findings (verified on master v0.0.49)

**Gap 1 (openai)** — the report pointed at the `imgs.length === 1 && firstParsed` sidecar condition, but current code has an additional filter upstream: `allImageParts` only collected image parts whose url parses as a data URL (introduced by 626f933 to feed `rawOpenaiContentParts`). A single remote-URL image therefore never reached the sidecar logic at all (`imgs.length === 0`). Same end state as reported, but the fix must also remove that filter. Reproducing also showed the gap is **wider than reported**: due to the same filter, multi-image messages dropped *every* remote-URL part — `[data, remote]` round-tripped as `[data]`, not just the single-image case.

**Gap 2 (responses)** — reproduced exactly as described: an image-only user item produced zero core messages. One nuance: loss was path-dependent — `patchResponsesInput` kept such items via the layout fallback (slot without coreId → original item), while `coreToResponses` dropped them; and in either path the untracked item stayed invisible to the compression pipeline and would accumulate unbounded. Tracking fixes both paths.

### openai codec (`src/wire/openai.ts`)
- `allImageParts` now collects every `image_url` part with a string url (data-URL filter removed).
- Single-image sidecar keeps the raw part verbatim for any URL shape; base64/mediaType fields are split out only for data URLs (the shape suggested in the issue):
  - 1 image → `{ rawOpenaiContent }` (+ base64 fields iff data URL)
  - >1 images → `{ rawOpenaiContentParts }` (field unchanged, now complete)
- `coreToOpenai` needed no change — its precedence chain already prefers verbatim raw over base64 rebuild.
- Image-only messages keep empty text (pre-existing behavior; switching to a placeholder would shift derived ids for already-working shapes).

### responses codec (`src/wire/responses.ts`)
- Image-only user items are now tracked: `"[image]"` placeholder text (anthropic codec precedent) + `rawResponsesItem` sidecar + data-URL base64 split as before, so they enter the compression space with stable ids.
- Multi-image-only items (≥2 `input_image`, no text) also get `"[image]"` — previously tracked with the `"\n"` join artifact, which is not real text. This is the **only id churn** introduced: those items' ids shift once (deterministically); every previously working shape keeps its exact id.
[...]

PR #292, opened 2026-09-14
fix(wire): preserve remote-URL and image-only images across the core-view boundary (#187, #291)

### LINKED ISSUE #291: [Bug] wire codecs lose images at the core-view boundary: openai remote-URL parts dropped (lost on every main-pipeline rebuild); responses image-only items dropped at toCore
来源: ranxianglei/billion-context#781 分析「折叠区间里的图片对摘要模型不可见」时发现（billion-context 侧的摘要占位修复见 PR ranxianglei/billion-context#786）

## Gap 1 — openaiToCore drops non-data-URL `image_url` parts (silent loss on rebuild)

`allImageParts` only sidecars image_url parts whose url parses via `parseDataUrl`. A **remote http(s) URL** gets no sidecar at all:

```
openaiToCore([{ role:"user", content:[{type:"text",text:"see"},{type:"image_url",image_url:{url:"https://example.com/chart.png"}}] }])
→ [{ id:"m00001", role:"user", contentType:"text", text:"see\n" }]   // no rawOpenaiContent(Parts), no imageBase64
```

Same input with a `data:` URL → `rawOpenaiContent` + `imageMediaType` + `imageBase64` set correctly.

**End-to-end proof (billion-context v0.1.111, acp-kernel 0.0.69):** openai `/v1/chat/completions` request with one remote-URL image part, main pipeline (`max_tokens: 4096`, compress tool injected → body rebuilt from `processedMessages`): the forwarded upstream body **no longer contains `https://example.com/chart.png`**. The model receives an image the client believes it sent, on every forward that goes through the kernel. (Side requests — `max_tokens <= SIDE_REQUEST_MAX_TOKENS` — do raw-body passthrough and keep the URL, which is why this only bites on the main pipeline.)

## Gap 2 — responsesToCore drops image-only items entirely

The "message" case gates on `if (effText)`: an item whose content is only `input_image` parts produces **zero** core messages:

```
responsesToCore([{ type:"message", role:"user", content:[{type:"input_image", image_url:"data:image/png;base64,..."}] }])
→ []   // nothing
```

Consequences: such turns are invisible to the kernel's token accounting and can never be folded; any rebuild-from-processed-messages path (e.g. billion-context `patchResponsesInput`, src/server.ts) also loses them. Mixed text+image items survive because `rawResponsesItem` keeps the full original item.

## Impact

Silent content loss toward the upstream model — no error, no log, host and user cannot detect it. Remote-hosted images (screenshots served over http, CDN assets) are the common openai case; image-only user turns (a bare screenshot reply) are common in interactive agents on the Responses API.

## Suggested fix

- `openaiToCore`: sidecar ALL `image_url` parts into `rawOpenaiContentParts` regardless of data:/remote scheme (use `parseDataUrl` only for filling `imageBase64`/`imageMediaType`). `coreToOpenai` already round-trips the parts array, so this closes the rebuild loss.
- `responsesToCore`: keep image-only items in the core view (e.g. a zero-text message carrying `rawResponsesItem`) instead of dropping them, so accounting/folding/rebuild all see them.

Both changes are additive to consumers of the current contract: billion-context #786 reads exactly these sidecars to render summary placeholders and works unchanged.

ranxianglei:
## What this fixes

Fixes #291 (exact duplicate of #187 — same two gaps, same root cause; see #291 thread for triage).

**Gap 1 — `openaiToCore` dropped non-data-URL `image_url` parts.** `allImageParts` filtered to parts whose url parses via `parseDataUrl`, so a single remote http(s) URL got *no* sidecar at all, and mixed `[data, remote]` multi-image messages kept only the data parts. Every main-pipeline forward that rebuilds the body from processed messages silently lost those images.

**Gap 2 — `responsesToCore` dropped image-only items entirely.** The "message" case gated tracking on `if (effText)`; an item whose content is only `input_image` parts produced zero core messages — invisible to token accounting, never foldable, lost on every rebuild-from-processed-messages path.

## Changes

- `src/wire/openai.ts`: `allImageParts` collects ALL string-url image parts (data: and remote alike); a single image always gets the `rawOpenaiContent` sidecar, with `imageBase64`/`imageMediaType` still only set for parseable data: URLs. `coreToOpenai` needed no change — it already round-trips the parts array verbatim.
- `src/wire/responses.ts`: new shared helper `userDisplayText(content, text)` returns the item's text, or the `"[image]"` placeholder (anthropic codec precedent) when a user item carries image parts but no real text part (≥2 image parts join into a truthy `"\n"` that is not text). Used on **both** sides: encode gate `if (effText || (role === "user" && hasImage))` + id/text derivation in `responsesToCore`, and the decode-side verbatim re-emit condition in `coreToResponses`. Both sides computing the same expression over the same raw item keeps the round-trip byte-for-byte locked by construction (adopts the #189 review finding about lockstep duplication). Image-only items are now tracked core messages carrying `rawResponsesItem`, so they enter accounting/folding and prune by id in `patchResponsesInput` instead of surviving forever via the untracked layout fallback.
- `src/wire/bili-message.ts`: doc updates for the widened sidecar contract.
- `tests/wire-bili-message-roundtrip.test.ts`: 10 regression tests (5d/5e/5f openai remote-URL single / image-only / mixed-order; 6b–6h responses image-only tracking + verbatim round-trip, patchResponsesInput survival **and** pruning, id stability/distinctness across turns, multi-image join artifact, whitespace-text guard, empty-text-part id-stability guard, malformed url).

This re-applies closed-unmerged PR #189 onto current master (v0.0.69 base), whose closure claim ("master already preserves images through wire round-trips") does not hold for remote URLs or image-only items — pre-fix repro output is in the #291 thread.

## Verification

- `npm run typecheck`, `npm test` (748/748 pass), `npm run build` — all green.
- Issue repros verified fixed: remote-URL image part now rides `rawOpenaiContent` and round-trips verbatim through `coreToOpenai`; image-only user items produce exactly one tracked `[image]` core message and re-emit verbatim.
- Stash-check: with the src fix reverted, 8/10 of the new tests fail on master (6f and 6h are guard tests pinning already-working shapes, not repros).
- Interaction audit: `stripHistoricalImages` (#215) operates on the rebuilt wire body and now actually sees remote-URL images (previously they vanished before it); `patchOriginalItem` preserves image parts when patching text; #190's `image_url:null` skip is intact (test 5c).
- Additive to billion-context #786's sidecar readers (`rawOpenaiContent`/`rawOpenaiContentParts`/`imageBase64`/`imageMediaType`/`rawResponsesItem`) — shapes unchanged, more values populated.

## Behavior-change notes

- One-time id shift on upgrade for responses user items with ≥2 images and no text part: previously tracked under the `"\n"` join artifact, now under `"[image]"`. Deterministic; ClusterCounter keeps identical items distinct.
- A user item with an empty-string `input_text` part plus an image keeps its pre-fix kernel text `"\n"` (joining two empty part texts yields a truthy join artifact, so it is **not** re-seeded to `[image]`); its derived id is byte-identical to master's for that shape (pinned by test 6h); verbatim round-trip still holds — no data loss.
- Out of scope (residuals noted in #187 floor 3): assistant-role image-only response items remain untracked; the mirror builder cannot express images.

## Review status

Automated second-agent review was unavailable in the execution environment (subagent spawn failed: provider model not found). The author ran two structured passes over the diff — internal consistency (id stability, encode/decode lockstep across all user-content shapes, assistant-path untouched, sidecar precedence) and contract/adversarial (consumer audit, edge cases, history interaction, README claims) — plus full verification above. Per AGENTS.md §6, please get at least two agent/human reviews before merging.


ranxianglei:

[bot] 🏷 **PR review — positive, one small issue fixed directly on the branch.**

**Diff cleanliness (first gate):** clean. Single commit on `origin/master` tip (4e6269d), 4 files, every line on-purpose (openai/responses/bili-message doc + tests). `mergeable_state: clean`, no unrelated churn. My review commit below adds only a 21-line test.

**What I verified (independently, not from the PR description):**

1. **Both gaps reproduce on un-fixed master** — re-ran the PR's test file against `origin/master`: exactly **8/9 fail** (5d/5e/5f Gap-1 remote-URL loss; 6b–6e/6g Gap-2 image-only drop); 6f passes as the guard test, matching the claimed signature. Also probed master directly: pure image-only item → 0 tracked messages; `[input_text "", input_image]` → tracked under text `"\n"`, id `h_28e0ad6df4a9e3f6`.
2. **Fix logic, both codecs:** `allImageParts` widening is correct — single image always gets `rawOpenaiContent`, base64/mediaType still data-only, multi keeps order (verified mixed `[data, remote]` round-trip verbatim). For responses, encode gate `effText || (role === "user" && hasImage)` + shared `userDisplayText()` on both sides holds lockstep *by construction*: `splitDemotedThinking` is assistant-gated so user `effText === messageContent(content)`, and each core message carries its own `rawResponsesItem`, so the decode comparison can't cross-contaminate even when a user literally types `"[image]"` (probed: distinct ids via ClusterCounter, both round-trip verbatim).
3. **Id stability:** the one-time shift is confined to the documented class (≥2 images, no text part: `"\n"` → `"[image]"`). All other working shapes keep byte-identical ids — spot-checked the empty-text+image shape above: identical id on master and PR.
4. **Pipeline participation:** image-only items now carry a `layout.coreId`, so `patchResponsesInput` patches them by id (`patchTextParts` keeps image parts when text changes) and prunes them when folded (test 6c); previously they survived forever via the untracked layout fallback. `stripHistoricalImages` (#215) is a pure function over the rebuilt wire body and now actually sees remote-URL images. Sidecar field audit across `src/`: only the own codecs read them — no hidden kernel dependency on the old data-URL-only values; additive to billion-context #786's readers as claimed.
5. **Green:** `npm run typecheck` ✓, `npm test` **748/748** ✓ (after my addition below), `npm run build` ✓.

**Small issue found & fixed directly (commit `f6a9179`, pushed to the PR branch):**

The behavior note claimed an empty-string `input_text` part + image "displays as `[image]` (the empty text is falsy)". That is wrong: joining `["", ""]` yields the truthy `"\n"` artifact, and the text part exists, so `userDisplayText` returns `"\n"` — I verified empirically on both branches. This is actually the *better* outcome: that shape already worked on master under `"\n"`, so keeping it avoids a second id shift (switching to `"[image]"` would move its id for zero functional gain — display/cosmetic only, since hosts render from sidecars). I chose to keep the code and pin reality: added **test 6h** locking `text === "\n"`, distinct-from-placeholder id, and verbatim round-trip for that shape, and corrected the PR body note via API (also 9→10 tests, 8/9→8/10 stash-check, 747→748). If you'd rather have display consistency (`"[image]"`) at the cost of that shape's id moving, say so and I'll flip it — but I recommend against it.

**Residuals (out of scope, correctly noted):** assistant-role image-only response items remain untracked; mirror builder can't express images.

**Status:** this counts as one structured review (correctness/id-stability + contract/consumer passes above). Per AGENTS.md §6 you need at least two independent agent/human reviews before merging — please get a second one (or confirm the prior two-pass self-review suffices in light of this). I can't merge PRs — system rules forbid it; please merge yourself: https://github.com/ranxianglei/acp-kernel/pull/292

中文摘要：复核确认两个图片丢失缺陷在 master 上真实存在且修复正确（远程 URL 走 sidecar 原样往返、纯图片 item 纳入压缩空间，id 稳定性除已声明的一次性迁移外逐字节不变），review 中发现并直接在分支上补了空文本+图片形状的锁定测试（f6a9179）并更正了 PR 描述中的一处行为误述；diff 干净、全量验证通过，满足规范后请安排第二人评审并由你合并。

PR #293, opened 2026-09-15
fix(prune,compress): never split a tool call/result pair across a fold

ranxianglei:
Cleaned-up, mergeable version of the fix for #286 — the same change as @joshwapohlmann's PR #287, with the incidental Prettier reformatting of `decideNudge` dropped so every changed line serves the fix.

## What this fixes (#286)
A fold covering part of a parallel tool burst can separate a surviving tool call from its result, producing an unrecoverable 400 on strict OpenAI-compatible backends ("an assistant message with 'tool_calls' must be followed by tool messages responding to each 'tool_call_id'"). Two independent holes:

1. **Summary anchor** — `collectSummaryAnchors`/`rebuildMessages` render the summary at `min(effectiveMessageIds)`; folding one pair of a burst drops it between a surviving sibling call and its result.
2. **Protected carve** — the soft-protection carve runs *after* `adjustBoundariesForToolPairs`; carving out exactly one half of a pair leaves an unanswered `tool_call_id` or an orphaned result.

## Changes
- `src/prune.ts`: new `pairSafeAnchorIndex()` advances a summary anchor past any surviving sibling's result (fixed-point), preserving #171 rendered-position semantics.
- `src/compress.ts`: `applySingleRange` now withdraws any call/result pair split across the fold (both halves stay visible), alongside the existing #684 turn-withdrawal; the error/warning text is generalized to cover both.
- `tests/tool-pair-integrity.test.ts`: three targeted tests (hole-1 anchor landing, hole-2 refuse-when-only-half-foldable, hole-2 withdraw+warn).

## Verification
- `tsc --noEmit` ✅ · build ✅ · `npm test` **741/741** ✅
- Minimal repro (interleaved burst, fold the inner pair) → post-fold projection keeps every call adjacent to its result (0 pairing violations).

## Relationship to #287
Same fix as @joshwapohlmann's #287, which I reproduced and verified end-to-end on v0.0.69. This branch lives in the main repo so it is directly mergeable without a fork rebase, and carries no unrelated formatting churn. Happy to close #287 in its favor.

All credit for the original diagnosis and fix goes to @joshwapohlmann.


ranxianglei:
ework 接单 review

ranxianglei:

Review done on the exact head (`fe26671`, branch `2026-09-15_tool-pair-clean`). Verdict: **mergeable as-is for #286**, with one scope finding about #287 you should decide on before closing it.

## 1. Diff cleanliness — pass

- Single commit on top of current master (`4e6269d`); no rebase needed. 3 files, +323/−3: `src/prune.ts`, `src/compress.ts`, `tests/tool-pair-integrity.test.ts`. No version bump, no lockfile churn.
- The previous blocker on #287 is resolved: `decideNudge` is now byte-identical to master. Every changed line in `src/compress.ts` sits inside the turn-integrity gate region (comment, pair-withdrawal loop, generalized error/warning text).
- Branch name and `fix(prune,compress):` commit message conform to convention.

## 2. Correctness — both holes fixed, logic verified line-by-line

- **Hole 1** (`pairSafeAnchorIndex`, src/prune.ts): fixed-point advance of the summary anchor past any surviving sibling call's result. I traced the interleaved-burst repro (`[callA, callB, resA, resB]`, fold inner pair): anchor moves to just past the burst, projection becomes `[callA, resA, SUMMARY]` with the orphaned half stripped downstream → 0 violations against the strict-provider rule. It also preserves #171 rendered-position semantics (forward-only, relative anchor order kept via stable re-sort in `rebuildMessages`). Complexity is bounded O(n²)-ish with n = message count — fine at context scale.
- **Hole 2** (pair withdrawal in `applySingleRange`, src/compress.ts): runs *after* the protected carve and the #684 turn gate, so it sees the final split state. Sides-differ test on `effectiveMessageIds` is correct in all four quadrants, and the fix is conservatively directional — worst case is over-withdrawal (more messages stay visible), never extra folding, so it cannot produce a new invalid wire shape. It also covers the reverse direction #286 mentions (both calls fold, results stay visible), since each pair's halves land on opposite sides. Generalized `splitDesc` error/warning wording is accurate.
- **Tests**: all three hit the right seams (hole-1 anchor landing asserted via `summaryIndex > grepResultIndex`; hole-2 refuse-when-only-half-foldable asserts the new error text; hole-2 withdraw+warn asserts block membership and post-fold visibility), and the local `unansweredToolCalls` validator mirrors the actual upstream rejection rule rather than a weaker proxy.

## 3. Gate — green locally and on CI

Ran on real Node v22.23.2 at the exact commit: `tsc --noEmit` ✅ · `npm test` **741/741** ✅ · `npm run build` ✅. GitHub CI on `fe26671`: `pr-validation`, `test (22)`, `test (24)` all success. This session is a fresh independent verification (separate from the earlier #287 review), satisfying the two-reviewer bar.

## 4. One finding — "same change as #287" is now stale (non-blocking for this PR)

#287's head moved to `dc12b3e` after it was first reviewed. It carries a **third piece #293 does not have**: a backward walk in `pairSafeAnchorIndex` that moves an anchor landing *inside* a consecutive assistant run back to the run's start (~40 lines in `src/prune.ts` + ~113 test lines). That piece addresses a different provider validation class (DeepSeek thinking-mode `reasoning_content … must be passed back to the API`), which is outside #286's stated scope — so #293 correctly fixes exactly what it claims, but closing #287 outright would drop that delta.

**Recommendation:** merge #293 as-is (it fully closes #286), keep #287 open for the reasoning-run anchor delta, and land that delta as a small follow-up PR on post-#293 master (or port it into #293 pre-merge if you'd rather do it in one shot — your call). Don't close #287 until that delta is somewhere.

## 5. Nit (no action taken)

The compress.ts pair loop keys off `contentType === "tool-call"` without the `role === "assistant"` check that the turn gate's `callIds` set and the prune-side anchor walker use. `CoreMessage` types role/contentType independently, but no wire shape produces a non-assistant tool-call, and even in that impossible case the behavior is conservative (withdrawal keeps things visible). Not worth churning a green, mergeable diff.

---

中文摘要：逐行审核确认 #293 干净可合——diff 无无关改动（`decideNudge` 已与 master 逐字节一致）、两处漏洞修复逻辑正确且方向保守、本地真实 Node 全量验证通过（typecheck/build/741 测试）且 CI 全绿；唯一待你拍板的是 #287 后来新增的"锚点回退到 assistant run 起点"补丁不在 #293 里，建议先合 #293、该增量另开小 PR 跟进，暂不关闭 #287。

ranxianglei:
麻烦一次改对吧 不要返工

ranxianglei:

收到，已一次改完 —— #287 里缺的第三块（assistant-run 锚点回退）已经补进本 PR，不再需要后续 PR。

## What landed

New commit `2f60bfb` on `2026-09-15_tool-pair-clean` (on top of `fe26671`): **fix(prune): never split an assistant run with a fold anchor** — exactly the delta from #287's head (`dc12b3e`), ported verbatim:

- `src/prune.ts` (+38/−10): `pairSafeAnchorIndex` now first walks the anchor **back** to the start of a consecutive assistant run before the existing forward tool-burst fixpoint runs. Rationale (in the expanded doc comment): consecutive assistant cores (reasoning run + text + tool calls) merge into one wire message, so a summary landing between them sends the calls to a strict-echo provider without their reasoning run — DeepSeek thinking mode rejects it (`The 'reasoning_content' in the thinking mode must be passed back to the API`).
- `tests/tool-pair-integrity.test.ts` (+113/−2): new `reasoning()`/`assistantText()` helpers, a `splitsAssistantRun()` validator mirroring the provider rule, a new test folding two read pairs out of a five-call turn with the reasoning run visible, and `splitsAssistantRun` guards added to the burst test.

Both files are **byte-identical to the reviewed fork head** (verified by hash/diff against `prfork/2026-09-14_tool-pair-integrity`), so nothing was re-derived by hand. The fork branch's `compress.ts` formatting drift (prettier re-wraps inside `decideNudge`) was deliberately **not** ported — it's unrelated churn, CI has no format job, and keeping this PR's `compress.ts` byte-identical-to-master-except-the-fix preserves the clean-diff property you asked for.

One behavior note (why the old burst-test assertion had to change): with the backward walk, in the interleaved-burst scenario the summary can now land at the **run start** (before the surviving call) instead of after the surviving result — both placements satisfy the provider rule, never the middle of a call/result span. The assertion was therefore relaxed from "after the result" to "not strictly inside the span", plus the new assistant-run guard; the forward-walk path is unchanged and still covered.

## Verification

- Local, real Node v22.23.2 at `2f60bfb`: `tsc --noEmit` ✅ · `npm test` **742/742** (was 741, +1 new test) ✅ · `npm run build` ✅
- GitHub CI on `2f60bfb`: `pr-validation`, `test (22)`, `test (24)`, `build-artifact` — all **success**.

## Net effect

#293 now covers all three provider-validation classes: turn integrity (#684), tool-pair splitting (#286), and assistant-run splitting (DeepSeek echo). Once merged, **#287 can be closed as fully superseded** — no remaining delta anywhere.

---

中文摘要：按你要求把 #287 独有的"折叠锚点不得落在连续 assistant 消息内部"补丁（含测试，逐字节取自已审查的 fork 头）直接补进了 #293 新提交 `2f60bfb`，本地 typecheck/build/742 测试全过、CI 四项全绿；合并本 PR 后 #287 可整体关闭，无需返工。

PR #313, opened 2026-09-16
fix(recommend): never emit sub-threshold tail batches from mergeRangesToThreshold (#309)

### LINKED ISSUE #309: mergeRangesToThreshold 输出低于 gate 阈值的尾部批次,破坏「列出即可压缩」不变量
来源: ranxianglei/billion-context#847 (https://github.com/ranxianglei/billion-context/issues/847) 分析 持续压缩失败、模型反复提交同一被拒范围 时发现

## 现象

billion-context #847:模型连续 7 次提交同一范围,每次都被 gate 拒绝(`Total compressible content too small (1734 chars across 1 range(s), min 5000)`),而 nudge/acp_status 全程把该范围列为可压缩(pendingT1 ≈ 40K 一直存在)。最终 usage 76%→89%,上游流截断。

## 根因

`mergeRangesToThreshold`(0.0.74 dist/index.js:2129-2145):把相邻 range 合并到 `batchChars >= minChars` 为止,但循环结束时**无条件 push 尾部残余批次**,即使其 `batchChars < minChars`:

```js
if (batch.length > 0) result.push(mergeBatch(batch));
```

因此 nudge 的 "Compressible ranges" 列表(`recommendNode` → `recommendedRanges`,dist/index.js:~2505-2540,注入点 :3083)会包含压不动的范围。**不变量「凡被列为可压缩的范围都一定压得动」被破坏。**

加剧因素:`viableRanges`(chunk-QPVKTDYF.js:387-390)的门槛是 `VIABLE_RANGE_MIN_TOKENS = 200` tokens(≈800 chars),与 gate 的 `compress.minCompressRange`(默认 5e3,**chars**)单位和数值都不一致——两个门槛、两种单位,消费方无法只靠一个过滤就对齐 gate。

## 影响

- 模型按列表选范围 → 必然被拒 → 失败调用+错误留在上下文尾部 → 模型模仿重试 → 死循环(#847 实测 7 次/13 分钟)。
- 所有依赖 `recommendedRanges` 的消费方(nudge、acp_status、preflight 式自动压缩)都被误导。

## 建议修复

1. 尾部批次若 `< minChars`:并入前一个批次(允许超阈值),或不输出。
2. 统一门槛:让 `viableRanges` 与 gate 共用同一阈值与单位(chars),或废弃双门槛之一。
3. 加一条回归测试:`recommendedRanges` 输出的每个 range 必须满足 gate 的 chars 下限(当前默认配置下必挂)。

ranxianglei:
## Context

billion-context #847: the model submitted the same range 7 times in a row, each rejected by the apply-side gate (`Total compressible content too small (1734 chars across 1 range(s), min 5000)`), while the nudge kept listing that range as compressible the whole time — a guaranteed-reject retry loop that drove usage 76%→89%.

## Root cause

`mergeRangesToThreshold` (src/recommend.ts) unconditionally pushed the residual tail after its flush loop, even when that tail's real char count was below `minCompressRange`. The sub-threshold batch flowed `recommendNode` → `NudgeDecision.compressibleRanges` → `formatRanges`, which prints every entry verbatim under "Compressible ranges" — no threshold check anywhere on the render path. Broken invariant: *every range listed as compressible must actually be compressible*.

(`pendingByTier` already encoded this invariant for tier-pending accounting — "so a nudge never offers a range the kernel would atomically reject" — but the range list itself never got the same treatment.)

## Fix

Enforce the invariant at the source: every batch returned by `mergeRangesToThreshold` now alone clears `minChars`:

- **sub-threshold tail → folded into the preceding flushed batch** (overshoot allowed; count/tokens/chars/userMsgs/dangerous aggregate correctly);
- **no preceding batch** (all content below the gate) → **nothing emitted** — the total is below the gate, so no selection of it can pass, and offering it only yields guaranteed-rejected calls.

This heals all consumers at once: kernel nudge text, the public `NudgeDecision.compressibleRanges` field hosts read directly, the panel, preflight-style auto-compress. The `pendingByTier` filter stays as defense-in-depth.

**On `viableRanges` (issue proposal #2):** `VIABLE_RANGE_MIN_TOKENS = 200` is left unchanged — it's a public export serving a different job (host-side display heuristic against "Summary too short" atomic rejection), and after this fix every recommended range ≥ minCompressRange chars trivially clears 200 tokens, so the unit mismatch becomes harmless on the recommend path. Happy to unify instead if you prefer.

## Tests

- New: `sub-threshold tail folds into preceding batch`, `all content below threshold → []`, plus an invariant test asserting every emitted batch alone clears `minChars` across shapes (incl. CJK-like real `chars`). All three verified FAILING on unfixed master.
- Updated 3 tests that asserted the old behavior: two in tests/recommend-merge.test.ts ("tail kept", "trailing flush"), one arbitration test in tests/nudge.test.ts — pendingT1 is now 2000 (folded m0..m9) instead of 1600 (filtered tail), and it additionally asserts the nudge lists exactly one gate-clearing range. tests/block-map.test.ts fixture raised above threshold so the userMsgs-summation assertion still exercises a mid-loop flush.

Pre-flight: typecheck ✓ · 790/790 tests ✓ · build ✓

Closes #309


ranxianglei:

[bot] 🏷 Verified this PR (agent-completed work per `ework-agent-pr` marker — reviewed, not redone).

**Diff review** — clean, 4 files, every line scoped to #309:
- `src/recommend.ts:352-355` — core fix confirmed: sub-threshold tail folds into the preceding flushed batch via `result[result.length - 1] = mergeBatch([prev, ...batch])`; if `result` is empty (all content below gate), nothing is emitted. Aggregation is correct on re-merge: `prev` carries a real `chars` field from its own `mergeBatch`, so `rangeChars(prev)` uses it (not the `tokens*4` fallback); `count`/`tokens`/`userMsgs` sum, `toolPct` stays a count-weighted mean, `dangerous` propagates, span extends to the tail's `endRef`. Doc comment updated to state the new invariant.
- Tests match the description: two old-behavior assertions rewritten (`tail kept` → fold-with-overshoot, `trailing flush` → `[]`), new invariant test covering clean flushes / foldable tails / CJK-like real `chars`, plus the nudge arbitration update (`pendingT1` 1600→2000, asserts exactly one gate-clearing range listed) and the block-map fixture bump.

**CI** (head `1f35417`): test (24) ✓ · test (22) ✓ · build-artifact ✓ · pr-validation ✓

**Local pre-flight on the branch**: typecheck ✓ · 790/790 tests ✓ · build ✓

Diff is merge-clean, invariant holds end-to-end (recommend → nudge text / `NudgeDecision.compressibleRanges` / panel / preflight auto-compress), `pendingByTier` kept as defense-in-depth. Ready for human merge.

---
中文摘要：已核对 PR #313 代码与 CI（全绿），本地复跑 typecheck + 790/790 测试 + build 全部通过；修复逻辑（低于阈值的尾部折叠进前一批次、无前置批次则不输出）与描述一致且聚合正确，diff 干净，可以合并。

PR #324, opened 2026-09-17
fix: order/merge nudge ranges by position, not ref number

ranxianglei:
## What

Fixes the kernel-side root cause of [billion-context#887](https://github.com/ranxianglei/billion-context/issues/887): when message ref numbers are non-monotonic vs current array order (subagent interleaving / mid-array summary nodes), `formatRanges()` merged candidate ranges by **ref number** while `resolveBoundaries()` resolves endpoints by **position**. A merged endpoint pair then collapses to a tiny slice, so the displayed range size no longer matches what is actually selectable and the `minCompressRange` gate fails with `Total compressible content too small`.

## Fix

Carry each range's first/last messages-array index (`startIndex`/`endIndex`) through `buildCompressibleRanges` -> `mergeBatch`, and order/merge by **position** (gap <= 1 slot) instead of ref number.

- Byte-identical output for dense monotonic refs (the common case).
- Hand-built ranges without indices fall back to the old ref-number behavior.
- Also eliminates over-fragmentation from ref holes (consistent with the codebase's "array adjacency, never ref arithmetic" philosophy).

Files: `src/types.ts`, `src/recommend.ts`, `src/nudge-text.ts`, + regression test `tests/range-coverage.test.ts`.

## Regression test

`tests/range-coverage.test.ts` reproduces #887's minimal scenario (5 messages, non-monotonic refs where a low-ref user message sits after higher refs) and asserts `resolve(format(ranges))` covers every original message. Verified it **fails on the pre-fix code** (`not ok 1`) and passes on the fix.

## Preflight

- `tsc --noEmit` — clean
- `node --import tsx --test tests/*.test.ts` — all pass (807 pass, 0 fail)
- `npm run build` — success

## Downstream (billion-context)

This is the kernel-side fix only. After this merges and CI publishes a new `acp-kernel`, `billion-context` must bump its pinned `acp-kernel` devDependency (currently `0.0.69`) to the new version in a separate change/release.


ranxianglei:

[bot] 🏷 Picking up this PR for verification. Initial read: diff is clean and focused (`src/nudge-text.ts`, `src/recommend.ts`, `src/types.ts`, + `tests/range-coverage.test.ts`; no version/lock churn). Note the branch base (`a679eca`, v0.0.75) is behind current master (`36f9a9e`, which includes #320 touching `recommend.ts`) — checking whether a rebase to current master stays green before giving a verdict. Will follow up with full results shortly.

PR #327, opened 2026-09-19
fix: name covering block in already-compressed gate; drop misleading block-ID retry hint

ranxianglei:
## Problem (observed live)

From https://github.com/ranxianglei/billion-context-pi/issues/470 (billion-context-pi 0.1.71, kernel 0.0.75):

1. `compress` of m08350–m08517 succeeds → b178 (T1, 97 msgs), warning lists 17 excluded protected refs.
2. Model re-targets m08459..m08517 → rejected: `Requested range(s) already compressed (e.g. m08459..m08517) … Current active blocks span b1..b178 — retry with startId/endId set to active block IDs in that span.`
3. Model runs `acp_status`, then **decompresses b178** — the post-compression rework the issue reports.

Reproduced deterministically against the installed 0.0.75 bundle (synthetic 30-message session: compress m00002..m00020 → b1, retry m00018..m00020): identical error text.

## Root cause

- The `minCompressRange` gate's "already compressed" variant carried a **blanket** hint to retry with block IDs. That advice is not actionable: a block-boundary range below the char threshold cannot fire, so the model either retries into the same rejection or escalates to `acp_status` + `decompress` of the block it just created.
- The gate message never named **which** block covers the requested refs, forcing a status round-trip to discover it.
- The protected-exclusion warning listed raw refs ("Excluded N protected message(s) m…, m…") with no guidance that those refs are intentionally left visible — which is exactly what induced the model to target them again.

## Why messaging, not resolution semantics

`activeOwnerAnchor` deliberately does not snap boundary refs into a T1 block's own direct messages (`src/boundaries.ts` doc comment): a silent snap would put the block's own summary inside the resolved range, marking it nested and consuming it into a **same-tier duplicate block**. This fix keeps that invariant and repairs only the diagnostic surface.

## Changes (`src/compress.ts`)

- New `coveringBlockIds()`: resolves each boundary ref (message ref via `messageRefs`, block ref expanded through `effectiveMessageIds` so a stale `bN` resolves to its higher-tier owner) to the active block(s) covering it.
- Gate "already compressed" message now names the covering block(s) and offers the actionable escape hatch: *use search_context or decompress bN if you need details from it*; stale-ref fallback wording when no active block covers the refs.
- Tier distillation/condensation advice is emitted **only when actually ready** (active T1 count ≥ `tier2Trigger`, or T2 ≥ `tier3Trigger`) with the concrete `bX..bY` span; the blanket span hint is removed.
- Protected-exclusion warning suffix: *they stay visible outside the new block; do not target them in another compress call.*
- Also fixes a latent crash this review surfaced: the gate previously read `consumedRanges[0]!` unconditionally — a TypeError on every too-small/all-unknown gate path (reproduced pre-fix).

## Verification

- New regression suite `tests/regression-bili-470.test.ts` (4 tests: naming + escape hatch + no stale hint; tier hint fires at threshold with exact span; exclusion-warning guidance; stale-ref fallback).
- Existing gate assertions updated to the new wording (3 sites in `tests/compress.test.ts`).
- `npm run typecheck` clean · `npm test` **828/828 pass** · `npm run build` success · touched files prettier-clean (repo-wide format drift on master is pre-existing and untouched here).

No §7.3 contract touched (diagnostic text only); no version bump (release branch territory). After merge + publish, billion-context-pi bumps its `acp-kernel` pin 0.0.75 → this version in its normal release flow.

## Addendum: re-analysis under lean mode (`promptPack: "lean"`)

Reporter confirmed the incident ran in lean mode. Re-analysis:

- Lean is a **surface-only** pack (`src/packs.ts` `leanPack`: one-line tool descriptions, philosophy/when-to-compress/tier/decompress sections nulled, condensed how-to-compress contract). It overrides prompts and tool chrome via `sanitizePackSurface` — pipeline config (protected-zone filtering, `preserveRecentMessages`, `minCompressRange`, boundary resolution, prune) is untouched. The incident transcript shows the identical kernel path reproduced above against the same 0.0.75 bundle that shipped `builtin:lean`. **Root cause unchanged.**
- Lean is an **amplifier**: it strips exactly the redundant guidance (full tool descriptions, when-to-compress/when-not sections, decompress philosophy) that could compensate for misleading diagnostic text in default mode. Under lean, at the moment of failure the model's entire ACP behavioral contract is ~8 short acpTags lines + one-line tool descriptions + the tool-result text itself — so a diagnostic-surface defect becomes a behavioral defect directly. Two lean acpTags lines even push the model to act on tool output ("if a ref is stale or missing, call acp_status … retry using the reported refs") — which the model followed — but nothing elsewhere in the lean surface said the excluded protected tail was *expected*, so the perceived "incomplete fold" had no contradicting signal, and the misleading block-ID hint became the dominant next action.
- This makes the messaging fix load-bearing precisely where lean leaves the model standing: the exclusion warning now forbids re-targeting the excluded refs ("they stay visible outside the new block; do not target them in another compress call"), and the gate error ends with "Nothing new to compress in that window. Continue the task." — an explicit stop signal replacing the block-ID retry pull.
- Process note: future diagnostic-wording changes should be reviewed against the lean surface, which has no chrome to fall back on.


ranxianglei:

[bot] 🏷 Verified PR #327 (`2026-09-19_fix-already-compressed-gate-guidance` @ d2d1c68, single commit on top of master e8900f6 / v0.0.77). Did not redo the work — pulled the branch and audited it against the description.

**Code review** — diff is exactly 3 files (+363/−15), matches the body point-for-point:
- New `coveringBlockIds()` (src/compress.ts:169) resolves each boundary ref — message ref via `messageRefs`, block ref expanded through `effectiveMessageIds` so a stale `bN` resolves to its higher-tier owner — to the active block(s) covering it.
- Gate "already compressed" message now names the covering block(s) and offers the actionable escape hatch (*use search_context or decompress bN if you need details from it*); stale-ref fallback when no active block covers the refs; the blanket *retry with startId/endId set to active block IDs* hint is gone.
- `tierActionHint()` (src/compress.ts:200) emits distillation/condensation advice only when actually ready (active T1 ≥ `tier2Trigger` / T2 ≥ `tier3Trigger`) with the concrete `bX..bY` span.
- Protected-exclusion warning suffix: *they stay visible outside the new block; do not target them in another compress call.*
- `consumedRanges[0]!` → `const firstConsumed = consumedRanges[0]`, still dereferenced only under the existing `consumedRanges.length > 0` guard.
- No §7.3 contract touched (diagnostic text only); the `activeOwnerAnchor` no-snap invariant is preserved — no resolution-semantics change.

**Local verification on d2d1c68**: `npm run typecheck` clean · `npm test` **828/828 pass** (incl. the 4 new tests in `tests/regression-bili-470.test.ts`) · `npm run build` success. Prettier: `src/compress.ts` and the new test file clean; `tests/compress.test.ts` fails `--check` on a single hunk (~line 556) that exists identically on master e8900f6 and is outside this PR's diff lines — pre-existing repo-wide drift, correctly left untouched here.

**CI**: all 4 checks green on d2d1c68 (build-artifact, pr-validation, test ×2).

**Two minor notes (non-blocking)**:
1. The body's claim that the gate "previously read `consumedRanges[0]!` unconditionally — a TypeError on every too-small/all-unknown gate path (reproduced pre-fix)" is not supported by the code: on v0.0.75–v0.0.77 and master, every such read sits inside the `consumedRanges.length > 0` branch of the ternary; the too-small path (final else) never touches index 0. The refactor itself is fine and clearer, but the "reproduced pre-fix" wording overstates what existed.
2. "touched files prettier-clean" — strictly, `tests/compress.test.ts` still fails `prettier --check` due to that pre-existing hunk elsewhere in the same file; the lines this PR touches are clean.

**Verdict**: correct, minimal, well-tested fix for the billion-context-pi#470 rework loop — the retry error is now actionable without a status round-trip or a decompress of the freshly created block. Recommend merging. Per repo rules I can't merge PRs — please merge yourself: https://github.com/ranxianglei/acp-kernel/pull/327 . After merge + publish, billion-context-pi bumps its `acp-kernel` pin 0.0.75 → this version in its normal release flow.

修复了 already-compressed gate 的误导性提示(不点名覆盖块、诱导模型对刚创建的块做 acp_status + decompress 返工):现在错误信息直接指明覆盖的块并给出 search_context/decompress 出口,层级提示仅在真正达到触发阈值时给出,排除警告也加上了"不要再对这些 ref 发起压缩"的指引;改动仅限诊断文本、未触碰任何 §7.3 契约,本地 typecheck/test(828/828)/build 全绿且 CI 全绿——可以合并(仅 PR 正文两处措辞小瑕疵,不阻塞)。

ranxianglei (commented):
Independent review (cross-checked from billion-context-pi#470) — **the kernel is the right layer; this PR is mergeable**.

Verification on my side:
- 828/828 pass on the PR branch; **merges cleanly onto current master (v0.0.80), 865/865 pass** on the merged tree — so a rebase (or merge as-is) is trivial.
- The diagnostics-only scope is the correct call: `activeOwnerAnchor` deliberately does not snap boundary refs into a T1 block's own messages (silent snap would fold the block's own summary into a new range → same-tier duplicate blocks). Fixing the *message* the gate emits, not the resolution semantics, removes the rework loop without touching compression behavior.
- `coveringBlockIds` handles both boundary kinds correctly: message refs resolve through `messageRefs.byRef` (with numeric fallback), block refs expand through `effectiveMessageIds` so a stale `bN` resolves to its higher-tier owner. Predicate now matches `danglingMessageRefs`' notion of coverage.
- `tierActionHint` only fires when the trigger count is actually reached — the old blanket "retry with startId/endId set to active block IDs" advice was unexecutable below the char threshold, which is exactly what sent the model into acp_status → decompress rework in the #470 transcript.
- The four regression tests in `regression-bili-470.test.ts` cover all three defect surfaces plus the no-covering-block fallback wording.

Answer to "有没有必要提到内核": yes — the root cause (boundary ownership lookup + gate error text) is entirely kernel-owned; the pi adapter only relays the kernel's error string, so a pi-side fix would be masking in the wrong layer. After this merges, the rollout needs the usual cross-repo sequence: kernel release → billion-context-pi pin bump.

